PR #247, opened 2026-05-09
feat(#187 AC#2): embedded LLM model downloader

jayzalowitz:
## Summary

The launch-gating piece. A non-technical user opening SkyTwin previously hit a wall: configure llama.cpp + GGUF + env vars, or buy API keys. With this PR: open Settings → "Local AI brain" card detects RAM bracket via `navigator.deviceMemory`, recommends the highest-quality model that fits, click Download → 3-15 minutes later the twin works fully offline. No API keys, no per-message cost, no manual install.

The registry shipped in #246 told us *which* model to download. This PR makes it actually happen — with pause / resume / cancel and persistence across API restarts.

## What ships

- `039-model-downloads.sql` migration + `modelDownloadRepository`
- `apps/api/src/embedded-llm/downloader.ts` — fetch + Range + atomic rename + SHA-256 verify + AbortController-based pause/cancel + boot-time orphan recovery
- `/api/embedded-llm/downloads/{start,:id,user/:userId,pause,resume,cancel}` endpoints
- `apps/web/public/js/components/embedded-llm-card.js` — Settings card, polls every 1s during active download, reuses `.confidence-bar` styles
- 14 new route tests; api total **477 passing** (24 skipped)

## Why this fits the theme

Pure boring-deterministic infrastructure: `fetch()`+Range, SHA-256, atomic rename. No LLM in the cryptography or the download path. Every byte accounted for; every state transition reflected in the row; every button has a deterministic backend consequence.

## Closes for #187

- ✅ AC#1 runtime
- ✅ AC#2 download with pause/resume
- ☐ AC#1 default GGUF in installer payload (separate distribution work)
- ✅ AC#3 STT runtime
- ☐ AC#4 Piper TTS (binary not installed)
- ✅ AC#5 model registry
- ☐ AC#6 mode-switch UI
- ✅ AC#7 same prompts
- AC#8 cost dashboard — implicitly satisfied

5 of 8 ACs closed. Remaining are distribution (AC#1 bundling), missing binary (AC#4), or small UI follow-ups.

## Test plan

- [x] `pnpm build` clean across all 34 packages
- [x] `pnpm --filter @skytwin/api test` — 477 passing
- [x] `vi.mock` ensures tests never touch real HTTP / filesystem
- [ ] CI green
- [ ] Manual: open Settings → click Download → watch progress

## Out of scope

- Real SHA-256 hashes (registry currently has placeholders; downloader skips verify when hash is all-zeros)
[...]

Issue #187, opened 2026-05-07
Capability loop #N: Embedded LLM with auto-upgrade path (Phi/Llama/Qwen via llama.cpp; Whisper-tiny STT; Piper TTS)


jayzalowitz:
**Parent epic:** #195
**Phase:** 3 — Reach
**Effort:** ~40h
**Depends on:** #185 (turnkey distribution to bundle the runtime)

## Context

The LLM-driven onboarding (#181), Twin Briefing (#177), reverse capability flow (#177), and adaptive judgment layer (#189) all silently degrade or break when no LLM provider is configured. Most non-technical users don't have an Anthropic or OpenAI API key. Either we ship an embedded model, or the magical features only work for technical users.

The architectural commitment: **the same prompts run against any model**. Better small-local models = better twin overnight, no code change.

## Scope

- Bundle Phi-3.5-mini OR Llama-3.2-3B OR Qwen2.5-3B via llama.cpp (final pick TBD based on benchmarking on M1/M2 + low-end Windows). ~2.5GB download.
- Background download post-install; user sees "your twin is downloading its brain — about 5 minutes" with progress bar; app remains usable in degraded mode during download.
- Whisper-tiny embedded for STT (~80MB).
- Piper embedded for TTS (~50MB).
- **Auto-upgrade path**: embedded model registry tracks "best small model that fits in this RAM bracket"; when a better model ships, user gets "your twin can be 30% smarter — download the upgrade?" prompt.
- **Mode switch**:
  - **Smart mode** (default for non-tech): embedded local model
  - **Smarter mode**: bring your own API key (Anthropic / OpenAI / Google) OR use SkyTwin Cloud (post-launch)
- `@skytwin/llm-client` already supports a provider chain — embedded model becomes a new provider type.

## Acceptance Criteria

1. Embedded model (Phi-3.5 / Llama / Qwen — final pick justified in commit message) bundled via llama.cpp; ~2.5GB download size; runs on 8GB RAM Macs and 8GB RAM Windows machines.
2. First launch: model downloads in background; app usable in degraded mode (no LLM-driven features) until download completes; progress bar shown; user can pause/resume.
3. Whisper-tiny embedded for STT; voice install (#179) works without internet.
4. Piper embedded for TTS; spoken briefing works without internet.
5. **Auto-upgrade**: embedded model registry checked weekly; new model available → user prompt; user-controlled (never auto-applies).
[...]

PR #248, opened 2026-05-09
feat(#187 follow-up): first-run dashboard "needs a brain" prompt

jayzalowitz:
## Summary

A tiny but launch-critical UX gap: a brand-new user lands on the dashboard with no AI provider configured and an empty state that offers no path forward. They're left to discover Settings → AI brain on their own.

This PR surfaces a banner when **zero AI providers are enabled AND zero decisions exist yet** — the natural first-run state. Two CTAs route to Settings:

- **Set up the local brain** — where the #187 AC#2 model downloader lives (PR #247)
- **Or bring your own API key** — for users who'd rather use a paid provider

Privacy framing leads: *"SkyTwin runs locally on your machine — no API keys, no per-message cost, your data never leaves the device. Pick up a free model in 5 minutes, or bring your own paid provider if you'd rather."*

That's the real first-run differentiator vs. ChatGPT-style assistants. The dominant first-run fear is "what is this thing going to do with my data?" — leading with local-first answers it before they have to ask.

## Why now

Once #247 (model downloader) lands, the local-brain path is fully self-serve in 5 minutes. But the dashboard didn't tell the user that path exists. This PR closes that gap and means the dashboard works as a first-run surface even before the user opens Settings.

Independent of #247 — this routes to Settings either way. Once #247 merges, the route lands on the model downloader card. Until then, it lands on the AI provider config (still self-serve, just paid).

## Implementation

- `apps/web/public/js/pages/dashboard.js` — new `renderBrainPrompt()`; gated by `enabledProviderCount === 0 && recentDecisions.length === 0` and `!tourMode` (seeded demo user has providers pre-configured)
- Reuses `GET /api/settings/:userId` via `fetchSettings` — **no new API endpoint**
- One new dashboard card, slotted between the "since last visit" banner and the Connect Google hero

## Test plan

- [x] `pnpm --filter @skytwin/web build` — clean
- [ ] New user (no providers, no decisions, not in tour) — banner appears
- [ ] Tour mode user — banner hidden
- [ ] User with 1+ enabled provider — banner hidden
- [ ] User with no providers but ≥1 decision — banner hidden (they got past first-run somehow)
- [ ] CTAs route to `#/settings`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,34 @@
 All notable changes to SkyTwin will be documented in this file.
 
+## [unreleased] — First-run dashboard "needs a brain" prompt (#187 follow-up)
+
+A tiny but launch-critical UX gap: a brand-new user lands on the dashboard
+with no AI provider configured and an empty state that offers no path
+forward. They're left to discover Settings → AI brain on their own.
+
+Closes the gap with a banner that surfaces only when (a) the dashboard
+fetch promises resolved successfully (no false-positive on transient API
+errors), (b) zero decisions exist yet, AND (c) zero AI providers are
+enabled. The provider check matches the Settings UI's truthiness rule
+(`p.enabled !== false`), so existing rows without an explicit `enabled`
+field aren't misread as off.
+
+Two CTAs: "Set up the local brain" (routes to Settings → Local AI brain
+card from #187 AC#2) and "Or bring your own API key". Privacy framing
+is per-option (not global): the local-first claim is scoped to the
+local brain row; the API-key row clearly notes "each message goes to
+that provider." Skipped in tour mode (seeded demo user has providers
+pre-configured), so the demo experience stays clean.
+
+Settings is fetched conditionally — only when the cheap prerequisites
+(`!tourMode && recentDecisions.length === 0`) already point at first-run.
+Once the user has any activity, no extra `/api/settings` round-trip per
+SSE-driven dashboard re-render.
+
+Implementation lives in `apps/web/public/js/pages/dashboard.js`. Reuses
+the existing `GET /api/settings/:userId` endpoint via `fetchSettings` —
+no new API. One new card, no other UI changes.
+
 ## [unreleased] — Embedded LLM model downloader (#187 AC#2)
 
 Closes AC#2 of #187. The single piece between "developer tool" and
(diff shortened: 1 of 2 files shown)


jayzalowitz:
Pushed [b00192b](https://github.com/jayzalowitz/skytwin/commit/b00192b) addressing all 3 Copilot findings:

**Privacy copy was misleading** ([line 253](https://github.com/jayzalowitz/skytwin/pull/248#discussion_r-2412857562)) — fixed. "Your data never leaves the device" was true only for the local brain path; the BYO API-key path obviously sends each message to the third-party. Restructured the card so the privacy claim is **scoped per-option**:

> **Local brain** — runs on your machine, no API keys, no per-message cost, your data never leaves the device.
> **API key** — uses Anthropic / OpenAI / Google. Faster on a small laptop, but each message goes to that provider.

Headline is now neutral: "Pick how your twin thinks."

**Stale provider state via slowFetch** ([line 299](https://github.com/jayzalowitz/skytwin/pull/248#discussion_r-2412857563)) — fixed. Dropped the `slowFetch` wrapper. The endpoint is small, this fetch is only relevant for first-run users (who don't have many other dashboard re-renders happening), and the cache savings weren't earned vs. the cost of "user enables a provider and the prompt sticks around for 30s."

**False positive on transient API failure** ([line 369](https://github.com/jayzalowitz/skytwin/pull/248#discussion_r-2412857564)) — fixed. Both `fetchSettings` and `fetchDecisions` falling back to empty arrays meant a single blip would surface the onboarding banner to users who actually have providers + decisions. `showBrainPrompt` now requires **positive evidence**:

```js
const showBrainPrompt = !tourMode
  && settingsFulfilled
  && decisionsFulfilled
  && enabledProviderCount === 0
  && recentDecisions.length === 0;
```

Web build clean.

jayzalowitz:
Pushed [c9fa927](https://github.com/jayzalowitz/skytwin/commit/c9fa927) addressing both findings from your re-review:

- **Conditional fetch** (line 306): `fetchSettings` removed from the Promise.allSettled batch. Now gated on the cheap prerequisites (`!tourMode && decisionsFulfilled && recentDecisions.length === 0`) — once the user has any activity, the fetch doesn't run at all. SSE-driven re-renders no longer hit /api/settings every tick.

- **Provider enablement convention** (line 382): aligned with `apps/web/public/js/pages/settings.js` (lines 1220, 1340). Now uses `p?.enabled !== false` so existing rows without an explicit `enabled` field are treated as on, matching the Settings UI's truthiness rule.

PR #249, opened 2026-05-09
fix(#187 AC#2 r3): multi-user partial collision + pause race + zombie polling

jayzalowitz:
## Summary

Follow-up to PR #247 (the embedded LLM model downloader). Copilot's third-round review landed after I merged — four substantive findings that need fixing.

### 1. Multi-user partial-file collision

Two users on the same API host downloading the same model both wrote to `<modelDir>/<modelId>.gguf.partial`. Concurrent streams would corrupt each other; one user's cancel would delete the other's partial.

**Fix**: the final GGUF is content-addressable (we SHA-256 verify before atomic rename), so it stays shared at `<modelDir>/<modelId>.gguf`. Partials now namespace by the download row id: `<modelDir>/<modelId>.gguf.<download.id>.partial`.

### 2. Pause-on-pending was silently overridden

`pauseDownload()` set DB to `paused`, but the already-scheduled `runDownload()` would start anyway and immediately overwrite back to `downloading`. Same path for pending→cancelled — the user's intent got nuked by the async runner picking up the pre-pause row.

**Fix**: at the top of `runDownload()`, re-fetch the row from DB. If status flipped to `paused`/`cancelled`/`complete`/`failed` between `startDownload()` returning and the runner picking up, return early.

### 3. Polling continued after navigating away

The 1s poll callback in `embedded-llm-card.js` had no termination condition tied to page navigation. Going to Approvals while a download was in flight kept hitting `/api/embedded-llm/downloads/:id` every second forever and held a reference to a detached `#embedded-llm-card-target` node.

**Fix**: poll callback now checks `window.location.hash !== '#/settings'` and `document.getElementById(CARD_TARGET_ID) !== container` at the top of each tick, and self-cancels.

### 4. `ensureDirectory()` dead code

Already removed in round-2 fix on PR #247 itself.

## Test plan

- [x] `pnpm build --concurrency=1` clean
- [x] 25 embedded-llm route tests still passing
- [ ] Manual: two simultaneous downloads (different users, same model) don't collide
- [ ] Manual: click Pause on a `pending` row before transfer starts → row stays `paused`
- [ ] Manual: start a download, navigate to Approvals → polling stops within 1s

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,37 @@
 All notable changes to SkyTwin will be documented in this file.
 
+## [unreleased] — Embedded LLM downloader: round-3 review fixes (#187 AC#2 follow-up)
+
+Copilot's third-round review of PR #247 landed after merge — four
+substantive findings, addressed in a follow-up PR:
+
+- **Multi-user partial-file collision**: two users on the same API
+  host downloading the same model both wrote to
+  `<modelDir>/<modelId>.gguf.partial`, so concurrent streams could
+  corrupt each other and one user's cancel could delete the other's
+  partial. The final GGUF is content-addressable (SHA-256 verified
+  before rename), so we keep that shared at `<modelDir>/<modelId>.gguf`,
+  but partials now namespace by download row id:
+  `<modelDir>/<modelId>.gguf.<download.id>.partial`.
+
+- **Pause-on-pending was a no-op**: `pauseDownload()` would set DB to
+  `paused`, but the already-scheduled `runDownload()` invocation
+  would start anyway and immediately overwrite back to `downloading`
+  via `setStatus`. Same bug for pending→cancelled. Fixed by
+  re-fetching the row at the top of `runDownload()` and bailing
+  early if status changed to `paused`/`cancelled`/`complete`/`failed`
+  between `startDownload()` returning and the runner picking up.
+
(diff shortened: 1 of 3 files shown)


jayzalowitz:
Pushed [f5e567e](https://github.com/jayzalowitz/skytwin/commit/f5e567e) addressing all 3 findings:

- **Install idempotency** ([downloader.ts:417](https://github.com/jayzalowitz/skytwin/pull/249#discussion_r-2412953812)): replaced the `unlink-then-rename` dance with: if target exists → discard our partial; else renameSync directly; if the rename throws AND target now exists (race lost) → treat as installed. No more way to destroy a known-good shared file.

- **Legacy partial migration** ([downloader.ts:60](https://github.com/jayzalowitz/skytwin/pull/249#discussion_r-2412953813)): added unconditional cleanup at the top of `runDownload()` — any pre-PR-#249 `<target>.partial` on disk is deleted before this row's namespaced partial is created. Avoids orphan junk after the upgrade.

- **Stale `cancelDownload` comment** ([downloader.ts:203](https://github.com/jayzalowitz/skytwin/pull/249#discussion_r-2412953814)): docstring updated to mention the namespaced format and the fact that the shared final GGUF is left alone.

Build clean. 25 route tests still passing.

Issue #251, opened 2026-05-11
Memory bootstrap: weight user-sent emails higher than received


jayzalowitz:
## Concept

When bootstrapping a user's memory layer (gbrain `brain_pages` and mempalace drawers) from a fresh Gmail/Calendar connect, emails the user **sent** are categorically higher-signal than emails they **received**, and the indexer should treat them that way. Today they're indexed at identical weight.

The principle is simple: **authorship = endorsement + intent**. The user typed the words, picked the recipient, decided the moment was worth the time. That's a fundamentally different epistemic claim than "this email arrived in their inbox."

This issue is about that asymmetry, what to do about it, and the (substantial) downstream effects it has on the rest of the system.

---

## Problem today

`apps/api/src/routes/events.ts:recordSignalToMemory` writes every email signal to `brain_pages` at the same weight. The Gmail connector at `packages/connectors/src/gmail-connector.ts` already knows whether a message has the `SENT` label, but the downstream memory layer doesn't use that information.

Three concrete failure modes follow from this:

1. **Bootstrap noise.** First-time-connect ingests N thousand inbox emails before it hits a single sent one (inbox > sent by orders of magnitude). Early semantic search hits, pattern inferences, and the dashboard's "what your twin knows about you" view are all shaped by Stripe receipts, AWS bills, and newsletter promos rather than by the user's own writing.
2. **Voice / style invisible to the twin.** A future `draft_email` candidate generator needs sent-corpus training data to write in the user's voice. Today that data is mixed in with inbox noise at 1:100 ratio — unrecoverable without a tier signal.
3. **Stated commitments are tier-less.** When the user writes "I'll have the doc by Friday," that's a first-person fact about a promise. Memory stores it next to "Promotional: limited time offer" with identical weight. The twin can't surface the commitment Friday afternoon.

---

## Why "authored" is structurally higher signal

Worth being explicit about the dimensions, because they each suggest different downstream uses:

| Dimension | Inbox tells you | Sent tells you |
|---|---|---|
| **Topical interest** | What others push at the user | What the user voluntarily writes about |
| **Style / voice** | How other people write | **How the user writes** — gold for `draft_email` |
| **Relationship hierarchy** | Who has the user's address | Who the user *chooses* to engage, and how (Hi vs. Dear, "thanks" vs. signed name, reply-in-5-min vs. reply-in-5-days) |
| **Commitments** | What was asked of the user | What the user **promised** — a first-person claim |
[...]

PR #252, opened 2026-05-11
feat(#251 Layer 1 + Layer 3 minimal): stamp authoring tier on email signals

jayzalowitz:
## Summary

First slice of #251. Every inbound Gmail signal is now classified into one of six `AuthoringTier` buckets at write time and stamped onto `RawSignal.data.authoringTier`. The embedded gbrain port projects that field onto `brain_pages.metadata` so the downstream retrieval layer can read it without joining back to the signal row.

**Closes (partial): #251 Layer 1 + minimal Layer 3.** Layer 2 (retrieval weighting) is intentionally NOT in this PR — that one is gated on `realistic-retrieval.test.ts` improving and ships separately.

## What changed

- `packages/connectors/src/authoring-tier.ts` (NEW) — `AuthoringTier` type + `classifyEmailAuthoringTier()` classifier. Six tiers: `user_sent_originated`, `user_sent_reply`, `inbox_personal`, `inbox_broadcast`, `inbox_newsletter`, `inbox_automated`. Field name is channel-agnostic so Slack/Notion connectors can extend the enum later without a schema change.
- `packages/connectors/src/gmail-connector.ts` —
  - Adds `To`, `Cc`, `In-Reply-To`, `List-Unsubscribe` to the `metadataHeaders=` URL on `messages/<id>?format=metadata`.
  - `messageToSignal()` calls the classifier and stamps `data.authoringTier` on every emitted `RawSignal`.
  - **Layer 3 minimal:** `bootstrapAndEmit()` now lists `in:sent newer_than:7d` BEFORE `is:unread newer_than:1d` and dedupes by id. First-impression brain pages lead with what the user wrote rather than what happened to be unread. Either list failing degrades gracefully to the other.
- `packages/memory-gbrain/src/embedded-port.ts` — `recordSignal()` reads `data.authoringTier` and projects it onto `brain_pages.metadata.authoringTier` when present. Falls back to the prior `{ signalSource, signalType }` shape when missing or non-string (defensive against future connectors).
- CHANGELOG entry explaining what shipped and what's deferred.

- **Layer 2 retrieval weighting.** The RRF fold and `DecisionMaker` pattern boost still read pages uniformly. Layer 2 lands as a separate PR after running labeled-relevant-doc evals — if R@5 doesn't improve, the weights don't ship.
- **Migration backfill.** Tier is only stamped on signals written after this lands. A backfill migration is cheap to write (re-read Gmail label + From header from stored signal rows) but is a separate concern from the live ingest path.

## Test plan

- [x] 18 classifier unit tests in `authoring-tier.test.ts` covering all 6 tiers, edge cases (empty input, malformed addresses, SENT-dominance, newsletter-beats-automated when both fire, To+Cc summed for broadcast threshold).
- [x] 6 new tests in `gmail-connector.test.ts` covering the end-to-end tier stamping path (SENT originated, SENT reply, newsletter via List-Unsubscribe, automated via noreply sender, broadcast via multi-recipient To, personal default).
- [x] 1 new test for sent-first bootstrap ordering (`bootstrap emits in:sent results before is:unread, deduped by id`).
- [x] 3 new tests for embedded-port metadata projection (string tier projected, missing tier omitted, non-string tier defensively dropped).
- [x] Full workspace test run: 70 tasks, all green.
- [x] Workspace build clean (`pnpm build --concurrency=1`).
- [x] No lint errors (`tsc --noEmit` in affected packages).

## Notes for reviewers

- The classifier is intentionally email-shaped today but uses channel-agnostic naming (`AuthoringTier`, `authoringTier`) so a future cross-channel reshape (`authored_originated` / `received_personal`) is a values change, not a schema change.
- The bootstrap dual-list adds one extra Gmail API call on first poll only; subsequent polls go through `history.list` exactly as before. Quota cost is two `messages?q=` calls + N `messages/<id>` detail fetches vs. one + N previously — negligible.
- No retrieval behavior change. `brain_pages.metadata` is JSONB; the new field is purely additive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,58 @@
 All notable changes to SkyTwin will be documented in this file.
 
+## [unreleased] — Memory bootstrap: stamp every signal with an authoring tier (#251 Layer 1)
+
+The first slice of #251. Layer 1 only labels the data — no retrieval-side
+weighting yet (that's Layer 2, gated on `realistic-retrieval.test.ts`
+improving) — so this PR is reversible and observable on its own.
+
+- **New: `AuthoringTier` enum + classifier in `@skytwin/connectors`.**
+  Six tiers (`user_sent_originated`, `user_sent_reply`, `inbox_personal`,
+  `inbox_broadcast`, `inbox_newsletter`, `inbox_automated`) capturing the
+  asymmetry between mail the user authored vs. merely received. The field
+  name is channel-agnostic on purpose (`authoringTier`, not
+  `gmailLabel`/`senderTier`); Slack/Notion connectors can extend the enum
+  later without rebuilding the memory schema.
+
+- **Gmail connector fetches the headers needed to classify.** Added
+  `To`, `Cc`, `In-Reply-To`, `List-Unsubscribe` to the
+  `metadataHeaders=` URL on `messages/<id>?format=metadata`. One extra
+  HTTP-header byte per message; the rest of the payload is unchanged.
+  Every emitted `RawSignal` carries `data.authoringTier`.
+
+- **Embedded gbrain port projects `authoringTier` onto
+  `brain_pages.metadata`.** When a connector stamps the field, the page
+  metadata gets `{ signalSource, signalType, authoringTier }`; when the
+  field is missing or non-string the page metadata falls back to the
+  previous two-key shape. No schema migration — `metadata` is JSONB.
+
+- **Layer 3 minimal: sent-first bootstrap ordering.** The first poll
+  for a new user now lists `in:sent newer_than:7d` BEFORE
+  `is:unread newer_than:1d` and deduplicates by message id. The user's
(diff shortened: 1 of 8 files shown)


jayzalowitz:
Round-2 review reply:

- **listMessageIds error swallowing**: Round-1 already changed this to propagate non-transient errors (commit cd9085d). Round-2's deeper point — that `fetchProfileHistoryId()` ALSO swallowed errors — is now addressed in fd302e2 (symmetric `RetryableHttpError` propagation).
- **AUTOMATED_DOMAIN_PATTERNS docstring**: round-1 already rewrote the docstring (cd9085d) to explicitly call out the deliberate asymmetry — `noreply\.` is intentionally NOT end-anchored to catch `noreply.<vendor>.com` subdomain aliases. The current comment block describes this; the round-2 flag appears to be reading the original text.

Issue #188, opened 2026-05-07
Capability loop #M: Turnkey distribution + embedded runtime (signed installers, single-binary, embedded SQLite, auto-update)


jayzalowitz:
**Parent epic:** #195
**Phase:** 3 — Reach
**Effort:** ~50h
**Depends on:** #173 (MCP host runs as part of the bundled binary), #191 (always-on background service)

## Context

The current epic assumes the user can run `pnpm dev` — a developer command. For "OSS launch for everyone" to be true, SkyTwin needs a turnkey installation: signed installers, no Node.js prerequisite, no Postgres prerequisite, single-binary distribution.

This child is pure infrastructure. No behavioral content. But without it, the OSS launch is "OSS launch for developers."

## Scope

- Mac App Store + Microsoft Store signed builds (electron-builder targets exist; this child wires the signing + store submission flows).
- Linux: Flatpak + AppImage signed.
- Auto-update via Sparkle (macOS) and Squirrel (Windows). Update channel: stable + beta.
- **Embedded SQLite single-user mode** as alternative to CockroachDB. Migration path: same schema, same repository layer; switch at config load. Multi-user installations keep CockroachDB.
- Single-binary installer with zero prerequisites: no Node.js install, no Postgres install, no MCP-runtime install (all bundled).
- First-launch experience: app launches, hits health check, ready in <10s.
- Telemetry opt-in (anonymous, opt-in only, transparent about what's sent).

## Acceptance Criteria

1. macOS `.dmg` signed + notarized, ships through Mac App Store + direct download.
2. Windows `.exe` signed, ships through Microsoft Store + direct download.
3. Linux Flatpak (Flathub) + AppImage signed.
4. Auto-update: launch the app on stable channel; update prompt fires when newer version is published; user can opt into beta channel.
5. **Embedded SQLite mode**: setting `SKYTWIN_DB_MODE=sqlite` swaps `@skytwin/db` to use SQLite via better-sqlite3. All migrations apply. All repositories pass their existing tests against SQLite.
6. First-launch from clean install on macOS: Finder → Applications → double-click → app open + health-check green within 10 seconds.
7. No prerequisites required: tested on a fresh macOS user account with no Homebrew, no Node.js, no Postgres.
8. Telemetry opt-in toggle in Settings; off by default; copy explains exactly what's sent.

## Out of Scope (this child)

- Auto-launch on system startup (handled by #191).
- Multi-user / multi-tenant installs (single-user is the default).
[...]

PR #253, opened 2026-05-11
feat(#187 AC#6 + AC#8): Smart/Smarter mode toggle + zero-cost helper

jayzalowitz:
## Summary

Closes the user-visible side of the embedded LLM story. Two pieces:

- **AC#6** — a two-pill Smart/Smarter mode toggle at the top of Settings → AI brain. Active mode is computed from the user's provider chain; clicking the inactive pill reorders priorities and auto-saves.
- **AC#8** — a `estimateLlmCostCents()` helper in `@skytwin/llm-client` that returns 0 for `embedded`/`ollama` and a per-provider rate for hosted APIs. The future spend-recording call site can drop this in without an embedded-special-case branch.

Together these take #187 from 5/8 → 7/8 closed. Remaining: AC#1 bundling (distribution work paired with #188) and AC#4 (Piper TTS, blocked on `piper` on PATH).

## AC#6 — Smart/Smarter mode toggle

**`detectAIMode(chain)` → `'smart' | 'smarter' | 'none'`** based on the top-priority enabled provider:
- `embedded` at top → Smart
- `anthropic` / `openai` / `google` / `ollama` at top → Smarter
- no enabled providers → none

**Click the inactive pill → priorities reorder + auto-save:**
- *Smart click:* `applySmartMode(chain)` promotes `embedded` to priority 0 (adds a fresh `embedded` entry with `model: 'auto'` if missing, re-enables a disabled one).
- *Smarter click:* `applySmarterMode(chain)` promotes the first hosted/Ollama entry to priority 0. Returns `null` if no candidate exists; in that case the action routes to `switch-to-smarter-blocked` which focuses the "+ Add a provider…" dropdown so the user's eye is drawn to the next step instead of failing silently.

After the save round-trip, `renderSettings` re-runs so the pill and the provider chain agree on the persisted state (handles cases like server-side normalization adding fields the optimistic copy didn't have).

**Pure helpers exported from `apps/web/public/js/pages/settings.js`** so the pill, the action handler, and any future audit route all agree on one definition. Same module-level `export function` style other helpers in this file use.

### Screenshot (Smart active, Anthropic configured as fallback)

`✓ Smart (free, on-device)` highlighted; `Smarter (paid API) →` is the call-to-action; helper text under each pill explains what they do.

## AC#8 — `estimateLlmCostCents()` helper

`packages/llm-client/src/cost.ts` exports:

```ts
estimateLlmCostCents(provider: AIProviderName, tokensIn?: number, tokensOut?: number): number
isZeroCostProvider(provider: AIProviderName): boolean
```

Rate table is keyed by provider with deci-cents per million tokens — list-price of the cheapest model we expose in `PROVIDER_MODELS` for each family. Local-runtime providers (`embedded`, `ollama`) carry zero per-token cost; that's the load-bearing piece of AC#8.

**Rounding is `ceil` everywhere** so spend-cap enforcement stays conservative — the failure direction is "approval required," never "silently past the cap."

**No call site wired yet.** The current spend-recording path (`spendRepository.create`) isn't called from any LLM code path today, so AC#8 is trivially satisfied at runtime. This PR adds the helper as a single source of truth so when LLM-cost recording does land (separate issue), it can compute `estimateLlmCostCents(response.provider, tokensIn, tokensOut)` and trust that local-runtime calls record zero.

## Test plan

- [x] 10 new vitest cases for `cost.ts`:
  - `embedded` returns 0 for any token volume (incl. `Number.MAX_SAFE_INTEGER`).
  - `ollama` returns 0.
  - Hosted providers compute expected integer cents with round-up.
  - Pure-function determinism check.
  - `isZeroCostProvider` distinguishes local-runtime from hosted.
- [x] 16 cases smoke-tested for the JS helpers via Node ESM import (`detectAIMode`, `applySmartMode`, `applySmarterMode` across empty/embedded-top/anthropic-top/ollama-top/all-disabled/lower-priority-embedded/skip-disabled scenarios + null-return when no candidate).
- [x] Toggle visually verified in Chrome at `http://localhost:3201/#/settings` across three scenarios:
  - Smart active (embedded at priority 0): Smart pill highlighted, Smarter pill is the CTA.
  - Smarter active (anthropic at priority 0): Smarter pill highlighted, Smart pill is the CTA.
  - No paid provider in chain: Smart pill clickable, Smarter pill routes to `switch-to-smarter-blocked`.
- [x] `pnpm --filter @skytwin/llm-client build` clean.
- [x] `pnpm --filter @skytwin/llm-client test` — 136/136 green.

## Notes for reviewers

- The mode toggle is a *layer* over the existing provider chain, not a replacement. Users who prefer the explicit chain editor still have it; the toggle is a single-click shortcut for the common cases.
- `applySmartMode` adding an embedded entry with `model: 'auto'` mirrors what `embedded-llm-card.js` uses for fresh installs — the runtime resolves the first GGUF in the detected modelDir.
- Cost rates are list prices reviewed at commit time. If pricing shifts and the table goes stale, the failure mode is over-estimating the bill (cap fires earlier) — never under-estimating. The unit test enforces `embedded === 0 AND ollama === 0` as an invariant.
- No JS test infrastructure exists in `apps/web`, so the helpers were smoke-tested via Node ESM import. If/when Playwright lands, the toggle's three-scenario behavior deserves an E2E.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,33 @@ All notable changes to SkyTwin will be documented in this file.
 
 ## [unreleased] — Smart / Smarter mode toggle + zero-cost helper (#187 AC#6 + AC#8)
 
+### Fixed (post-Copilot review)
+
+- Cost rate table was off by 100× — the original draft stored
+  `{ input: 8, output: 40 }` for Anthropic and called it "deci-cents
+  per 1M" when the conversion is actually `$0.80 → 80¢ → 800
+  deci-cents`. The table now stores `{ input: 800, output: 4000 }`
+  for Anthropic and the corresponding corrected values for OpenAI and
+  Google. The unit tests now pin the expected dollar-equivalent
+  outputs ($4.80 for 1M+1M Anthropic, $0.60 for 1M output OpenAI,
+  $0.30 for 1M output Google) so the unit conversion can't silently
+  regress again. Embedded + Ollama still return 0¢ at any volume.
+- `PUT /api/settings/:userId/ai` now accepts `embedded` as a valid
+  provider. The Smart mode toggle inserts an `embedded` entry; the
+  pre-existing validation set only allowed `anthropic` / `openai`
+  / `google` / `ollama`, so clicking "Use Smart mode" round-tripped
+  through `applySmartMode` and then 400'd at the API.
+- `switchAIBrainMode` rolls back the optimistic UI state on save
+  failure. Previously the pill + provider chain stayed in the
+  reordered state with only an error banner — visually implying the
+  switch succeeded when the server actually rejected it.
+- "Smarter" pill copy now says "paid API or Ollama" — the
+  `SMARTER_PROVIDERS` set includes Ollama (local, free) and the
+  earlier copy would have confused users running Ollama into thinking
+  the option didn't apply to them.
+
+### Original change
+
 Two pieces close out the user-visible side of the embedded LLM story:
 
 - **AC#6 — Smart / Smarter mode toggle in the AI brain card.** A two-pill
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,43 @@
 All notable changes to SkyTwin will be documented in this file.
 
+## [unreleased] — Smart / Smarter mode toggle + zero-cost helper (#187 AC#6 + AC#8)
+
+Two pieces close out the user-visible side of the embedded LLM story:
+
+- **AC#6 — Smart / Smarter mode toggle in the AI brain card.** A two-pill
+  row at the top of Settings → AI brain shows the user which mode is
+  active (computed from their provider chain: top-priority enabled =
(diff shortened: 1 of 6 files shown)


jayzalowitz:
Round-2 review reply:

**Already fixed in round 1 (efc0822):**
- Cost rates 100×-too-low → corrected to 800/4000 (Anthropic), 150/600 (OpenAI), 75/300 (Google) deci-cents per 1M tokens.
- Test expectations updated to pin the dollar-equivalent integer cents (.80 for 1M+1M Anthropic, etc.).
- Smarter pill copy updated to mention Ollama ("paid API or Ollama").

**Round-2 findings fixed in 0d20331:**
- `estimateLlmCostCents` throws on unknown provider (was: silent 0).
- Token-count safe-integer guard added (throws on non-finite, negative, or > 2e12).
- `applySmarterMode` docstring corrected to match behavior.
- `switchAIBrainMode` catch block now uses defensive `err instanceof Error ? err.message : String(err)`.

Issue #179, opened 2026-05-06
Capability loop #G: Mobile parity (Capabilities + Briefing + voice + push)


jayzalowitz:
**Parent epic:** #195
**Phase:** 3 — Reach
**Effort:** ~50h
**Depends on:** #176 (web UX as reference), #177 (briefing surface to mirror)

## Context

The mobile twin has to feel like the same twin. Suggestions arrive as push notifications when confidence is high; the user approves with a tap; voice-installs work via the existing speech-to-text path. No mobile-specific OAuth dance — install for stdio servers happens server-side; mobile is just remote control.

## Scope

- New screen `apps/mobile/src/screens/CapabilitiesScreen.tsx`: list suggestions (with evidence), approve/dismiss/snooze, view installed list.
- New screen `apps/mobile/src/screens/CapabilityDetailScreen.tsx`: per-server view, policy edit, regret button.
- New screen `apps/mobile/src/screens/BriefingScreen.tsx`: full briefing render with markdown.
- Update `apps/mobile/src/navigation.tsx` to add Capabilities + Briefing tabs.
- Voice trigger in `apps/mobile/src/voice/capability-voice-commands.ts`: "Hey twin, connect Linear" via existing speech-to-text → install via remote.
- Push notification in `apps/mobile/src/push/capability-push.ts`: high-confidence suggestion (`confidence_score >= 0.85`) → push notification. Hard rate-limit 2/day per user; user-disable per server.

## Acceptance Criteria

1. `CapabilitiesScreen.tsx`: list suggestions (with evidence), approve/dismiss/snooze, view installed list.
2. `CapabilityDetailScreen.tsx`: per-server view, policy edit, regret button.
3. `BriefingScreen.tsx`: full briefing render with markdown.
4. Voice trigger: "Hey twin, connect Linear" via existing speech-to-text → install via remote.
5. Push notification on high-confidence suggestion (`confidence_score >= 0.85`).
6. Reuses existing pairing/SSE/notification plumbing — no new mobile-side OAuth dance; install for stdio servers happens server-side.
7. Tests: ≥3 E2E (suggestion approve, briefing read, voice install).

## Out of Scope (this child)

- Initiating an OAuth connection from the device. Mobile shows the suggestion + approves it; the actual OAuth flow happens server-side and the user can complete it from web/desktop. Acceptable v1 limitation.
- Idle filesystem mining on mobile. Out indefinitely (the parent epic already gates this to desktop only).

## Files Reference

[...]

PR #254, opened 2026-05-11
feat(#179 voice): mobile voice recording + transcribe pipeline

jayzalowitz:
## Summary

Closes the **code-bound** half of #179. The mobile app can now capture audio and ship it to the paired desktop's `/api/voice/transcribe` (the route landed in PR #244). The remaining work on #179 is QA on physical iOS/Android devices, which has always been the actual blocker — see the [2026-05-11 status comment on #179](https://github.com/jayzalowitz/skytwin/issues/179#issuecomment-4419892942) for the corrected scoping.

## What changed

- **New `VoiceScreen` tab** (`apps/mobile/src/screens/VoiceScreen.tsx`). Tap-to-record → tap-to-stop → upload → transcript. Six-state machine (`idle` / `denied` / `recording` / `processing` / `result` / `error`) with the recording lifecycle driven by `useAudioRecorder` from `expo-audio`. Pulse animation + tabular-numeric timer while recording. Permission-denied state has a "How to fix" affordance pointing the user at system settings.

- **New `voice-service.ts` pure helpers**:
  - `audioFileToBase64(uri)` reads the recorder's output via `expo-file-system`'s `File.base64()` API (the SDK 55 idiomatic path).
  - `transcribeRecording(client, userId, uri, language?)` orchestrates base64 → upload → result mapping with **stable error codes** (`no_audio` / `read_failed` / `whisper_unavailable` / `network` / `unknown`) so the UI can branch on cause without parsing free-form error strings.

- **New `transcribeVoice(...)` on the API client**. Uses the existing `request<T>()` layer with a **60s timeout override** because whisper's first-run model load can take several seconds on cold start — the default 10s would abort mid-transcribe. New `TranscribeResponse` interface added to the response-type block.

- **Permissions in `app.json`**: `NSMicrophoneUsageDescription` (iOS), `RECORD_AUDIO` (Android), `expo-audio` plugin entry. Permission copy emphasizes "sent to your paired SkyTwin desktop for on-device transcription" so the install prompt matches the privacy story.

- **Tab nav**: `voice` added to the `MainTab` enum in `App.tsx`, wired into the `renderContent` switch, and a "Voice" `TabButton` placed between Capabilities and Dashboard.

- **Deps**: `expo-audio: ~55.0.14`, `expo-file-system: ~55.0.19` (versions match Expo SDK 55's `bundledNativeModules.json`). The latter was already transitively installed via `expo-asset`; declared explicitly so the dep is auditable.

- **TTS playback.** The screen displays the transcript but does not speak responses back. That's a separate flow that pairs with #187 AC#4 (desktop Piper TTS) once a `piper` binary is on PATH.
- **"Send to twin" hand-off.** The transcript is shown but not yet pipelined to the assistant or the decision route. A follow-up will route the transcribed text through the existing assistant route once that mobile surface lands.
- **Physical-device QA.** The Expo SDK 55 `expo-audio` API works in the simulator but real-device behavior (silent-mode switch, AirPods routing, background record interruption) needs hardware to verify.

## Test plan

- [x] **11 new vitest cases** in `voice-service.test.ts` covering:
  - `audioFileToBase64`: null/empty URI → `no_audio`, happy path, empty result → `no_audio`, `File.base64()` rejection → `read_failed`.
  - `transcribeRecording`: 200 happy path, language code forwarding, 503 → `whisper_unavailable`, 413 → `unknown` (with detail in message), missing URI short-circuits before `fetch` is called, network failure → `network`.
- [x] The test file mirrors the inlined-class pattern at `apps/mobile/src/__tests__/api-client.test.ts:23` (a `TestApiClient` stub instead of importing the real one) to keep React Native imports out of Node's test runner. Same convention already established for this app.
- [x] **Full mobile suite**: 165 tests, 163 passing + 2 skipped (the 2 skips are pre-existing in `integration-live.test.ts` — discovery tests, unrelated).
- [x] **Full workspace**: 70/70 turbo tasks green (`pnpm test`).
- [x] **`pnpm build --concurrency=1`** clean.
- [x] `npx tsc --noEmit` clean for `VoiceScreen.tsx`, `voice-service.ts`, `api-client.ts`, `App.tsx` (pre-existing TS errors in `discovery.ts`, `notifications.ts`, and `api-client.test.ts` are unrelated to this PR).

## Notes for reviewers

- The recording lifecycle lives inside `VoiceScreen.tsx` rather than in the service module because `useAudioRecorder` is a React hook — pulling it into a plain module would lose the hook's internal teardown and recording-state mirroring. The pieces that *are* easy to unit-test outside React (base64 conversion + transcribe orchestration with error mapping) live in `voice-service.ts` and are exercised in the test file.
- `expo-audio` is the SDK 55 path (`expo-av` was removed in SDK 53+). The hook API works cleanly here because the screen owns the recorder.
- Error codes are deliberately stable strings rather than free-form messages so a future analytics layer or audit-log surface can attribute voice failures without regex-parsing strings.


jayzalowitz:
Round-2 review reply: all 4 findings were already addressed in round 1 (commit 05e0449), but Copilot's round-2 read appears to be against the pre-fix state.

Current state on this branch:
- **type-only `SkyTwinApiClient` import**: already `import type` at line 19 of voice-service.ts.
- **`permission_denied` docstring**: already removed; the comment now lists the actual codes in the union (`no_audio`, `read_failed`, `whisper_unavailable`, `network`, `unknown`).
- **25MB base64 vs decoded**: docstring corrected to "25MB *decoded* (~33MB base64)" (line 261).
- **`openSystemSettings` Alert vs deep-link**: already wired to `Linking.openSettings()` with the alert as a fallback (commit 05e0449).
- **"bounds clip length" comment**: already rewritten to "The mobile recorder has no explicit cap; the screen surfaces the 413 if the user records past the limit" — no longer implies a non-existent guardrail.

PR #255, opened 2026-05-11
feat(#187 AC#4): Piper TTS backend + /api/voice/synthesize route

jayzalowitz:
## Summary

Closes **#187 AC#4**. Mirrors the proven spawn pattern of `LlamaCppTextBackend` and `WhisperCppSttBackend` — same shape, same diagnostics, same `Null*` fallback when the binary isn't installed.

Takes #187 from **7/8 → 8/8 closed** for code-bound ACs. The last remaining AC (#1 bundling) is distribution work paired with #188.

### New backend (`@skytwin/embedded-llm`)

- **`piper-tts-backend.ts`** — `PiperTtsBackend implements EmbeddedTtsPort`. Spawns `piper --model <model.onnx> --output_file <tmp> --quiet`, writes text to stdin, reads the resulting WAV into a `Buffer` on successful exit. Cleans up the tempdir on both success and failure. Bounded inputs: text required, max 8000 chars; mismatched voice request fails hard rather than silently substituting.
- **`findFirstPiperModel(dir)`** locates the first `.onnx` voice model with a paired `.onnx.json` config (Piper requires both). The pairing check is what differentiates a usable voice from a stray `.onnx` someone dropped in — catching it at detection keeps the failure visible at boot rather than at synth time.
- **`factory.ts → createEmbeddedTtsPort(overrides?)`** mirrors `createEmbeddedSttPort`: probe `runtime-detector` for a `piper` binary (env-var override → PATH lookup), then resolve a voice model (env-var override → first valid pair in the configured model dir). Falls back to `NullEmbeddedTtsPort` when either is missing.

### New API consumer (`apps/api/src/routes/voice.ts`)

- **`POST /api/voice/synthesize`** — body `{ userId, text, voice? }` → response `{ audioBase64, durationBytes, voice }`. Base64 instead of binary so it goes through the same JSON envelope the rest of the API uses (the mobile client + web dashboard both decode base64 → Blob/audio element).
- **503 + hint** when no piper binary is on PATH, matching the STT path's shape. Same recovery message (`install piper-tts and an .onnx voice model, or set SKYTWIN_PIPER_BIN + SKYTWIN_PIPER_MODEL`).
- **`GET /api/voice/capabilities`** now reports `stt` + `tts` capability blocks alongside the legacy STT-shaped fields so older clients keep working.

- **Bundled piper binary / voice model.** Still requires the operator to install piper-tts (`brew install piper-tts` on macOS, `apt install piper-tts` on Ubuntu) and drop an `.onnx` + matching `.onnx.json` config in the configured model dir. Bundling joins the same distribution work as #187 AC#1 (default GGUF) paired with #188 turnkey distribution.
- **Briefing → speech wiring.** The API surface is reachable now but the dashboard / mobile briefing screen doesn't yet auto-speak the current briefing. That's a UI follow-up gated on the existing briefing surface; the backend it needs is in main as of this PR.

## Test plan

- [x] **15 new vitest cases** in `piper-tts-backend.test.ts` covering:
  - Capabilities (`available=true`, voice list).
  - Happy path: stdin write + spawn args + WAV read + tempdir cleanup.
  - Non-zero exit: stderr tail surfaced in error message + cleanup still happens.
  - Exit 0 but missing WAV → rejection.
  - Exit 0 but empty WAV → rejection.
  - Empty text → synchronous reject, never spawns.
  - Text > 8000 chars → synchronous reject, never spawns.
  - Voice mismatch → reject, never spawns.
  - Voice match → resolves.
  - `findFirstPiperModel`: null dir, missing dir, paired-config check, non-`.onnx` skipped, orphan `.onnx` (no config) skipped.
- [x] **9 new API tests** in `voice-routes.test.ts` covering:
  - Updated `/capabilities` response shape (legacy + new `stt`/`tts` blocks).
  - `/synthesize` happy path (base64 WAV + voice in response).
  - Voice option forwarded to port.
  - 503 when piper unavailable.
  - 400 for missing userId / empty text.
  - 413 for text exceeding the 8000-char ceiling.
- [x] All mocks use `node:child_process` + `node:fs` stubs so tests run hermetically with no piper on the host.
- [x] **Full workspace**: 70/70 turbo tasks green (`pnpm test`).
- [x] **`pnpm build --concurrency=1`** clean.
- [x] `@skytwin/embedded-llm`: 86 tests passing (+15 new). `@skytwin/api`: 551 tests passing (+9 new).

## Notes for reviewers

- The `--quiet` flag on piper suppresses its banner + progress output on stderr so the diagnostics tail we collect on failure isn't 90% noise — same hygiene as the whisper-cli `-np -nt` flags.
- Voice mismatch hard-fails rather than silently substituting. Re-instantiating the backend with a matching `modelPath` is the documented way to switch voices.
- The `stt`/`tts` nested blocks in `/capabilities` are additive; legacy clients that read `body.available` / `body.supportedFormats` keep working unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ All notable changes to SkyTwin will be documented in this file.
 ### Original change
 
 Closes #187 AC#4. Mirrors the proven spawn pattern of
-`LlamaCppTextBackend` and `WhisperCppSttBackend`. Three pieces:
+`LlamaCppTextBackend` and `WhisperCppSttBackend`. Four pieces:
 
 - **`packages/embedded-llm/src/piper-tts-backend.ts`** —
   `PiperTtsBackend` implements `EmbeddedTtsPort`. Spawns `piper`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,28 @@ All notable changes to SkyTwin will be documented in this file.
 
 ## [unreleased] — Piper TTS backend + `/api/voice/synthesize` route (#187 AC#4)
 
+### Fixed (post-Copilot review)
+
+- `/api/voice` now mounts through `requireOwnership` so POST
+  `/transcribe` and `/synthesize` can't be called with another
+  user's `userId` in the body. The in-router
+  `bindUserIdParamOwnership` only covered `:userId` path params,
+  leaving body-userId POSTs structurally unprotected.
+- `PiperTtsBackend.spawnPiper` switched stdout to `ignore` (was
+  `pipe`). The WAV is read from `--output_file`, not stdout —
+  leaving stdout piped without consuming it could block Piper
+  once the OS pipe buffer filled. Matches the proven whisper-cli
+  spawn pattern.
+- Piper stdin now gets a trailing `\n` so the newline-delimited
+  reader treats the input as one complete utterance instead of
+  relying on EOF semantics. Matches the docstring; test updated.
+- `/api/voice/synthesize` response field renamed from
+  `durationBytes` to `audioBytes`. The former implied a time
+  measurement; the value is a byte count of the WAV. New endpoint,
+  no compat concern.
+
+### Original change
+
 Closes #187 AC#4. Mirrors the proven spawn pattern of
 `LlamaCppTextBackend` and `WhisperCppSttBackend`. Three pieces:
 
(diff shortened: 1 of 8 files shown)


jayzalowitz:
Round-2 review reply:

**Already fixed in round 1 (037938c):**
- `/api/voice` ownership: mounted with `requireOwnership` (apps/api/src/index.ts:242). Body `userId` is enforced.

**Round-2 findings fixed in 4ffac42:**
- `getPort()` renamed to `getSttPort()` to disambiguate from the new TTS port.
- "single port" wording in the header doc replaced — there are two ports now.
- CHANGELOG "Three pieces:" → "Four pieces:" to match the actual list.

Issue #193, opened 2026-05-07
Epic: Emergent Lifebooks — life-management surfaces that adapt to who the user actually is


jayzalowitz:
# Epic: Emergent Lifebooks — life-management surfaces that adapt to who the user actually is

> Companion epic to #195. Builds on the three-port architecture (`docs/architecture-philosophy.md`) to deliver life-management *verticals* without hardcoding any taxonomy.

## Context

After Epic #195 ships, SkyTwin has the substrate (capabilities, judgment, memory) but not the *life-management surfaces* — the dedicated views for Health, Money, Relationships, Time, Learning, Projects, Home that turn a tool into a life-manager.

The naive approach: hardcode a 7-vertical taxonomy with fixed memory schemas + curated capability bundles per vertical. **This is wrong.** It assumes every user's life decomposes the same way. What about Kayaking, Aging Parents, Job Search, Pets, Volunteering, Caregiving? Hardcoding seven verticals excludes everyone whose life doesn't fit.

The right approach: **emergent verticals**. The twin reads MemPalace, identifies what life domains matter to *this* user, creates dynamic memory wings, and renders per-domain surfaces. Some users get a Kayaking wing. Some get Caregiving. None of these are pre-baked.

Adding a new "kind" of Lifebook doesn't require a code deploy — just a better domain-extraction prompt.

## Architectural commitment

This epic strictly follows `docs/architecture-philosophy.md`:

- **Hard rails**: no Lifebook can bypass spend caps, FS denylist, or audit log
- **Boring deterministic**: domain wings are stored in the same `KnowledgeEntity` + wing/room/drawer schema MemPalace already provides; no new vertical-specific tables
- **Adaptive**: domain extraction, per-Lifebook briefings, capability suggestions per domain — all prompt-driven via `@skytwin/policy-prompts` (#187)
- **Memory port**: all reads/writes through `@skytwin/memory-port` (#191); supports any backend that exposes spatial wings

## Architecture

```
                ┌─────────────────────────────┐
                │ MemPalace via memory-port   │
                │ (signals + entities + facts)│
                └─────────────┬───────────────┘
                              │
                              ▼
              ┌──────────────────────────────┐
              │ domain-extraction prompt      │
              │ (from @skytwin/policy-prompts)│
              │  weekly worker, per user      │
              └─────────────┬─────────────────┘
                            │ produces
                            ▼
              ┌──────────────────────────────┐
              │ [{ domainName, importance,    │
[...]

PR #256, opened 2026-05-11
feat(#193 follow-up): filter the capabilities registry by Lifebook

jayzalowitz:
## Summary

Closes the **"capability filter by domain on Capabilities page"** item that PR #242 (#193 Child 1) explicitly deferred. The Capabilities page now carries a Lifebook dropdown alongside the existing category filter; selecting a Lifebook intersects the registry results with that Lifebook's `suggestedCapabilities: registryId[]` set.

## What changed

**One file: `apps/web/public/js/pages/capabilities.js`**

- Fetch `/api/lifebooks/:userId` alongside the existing capabilities + recipes + opt-ins data on render. Best-effort — failure degrades to "no Lifebook filter offered" rather than blocking the page.
- New `renderLifebookFilterDropdown(lifebooks)` renders the `<select>` only when at least one visible Lifebook exists; hidden Lifebooks (`hiddenAt` set) are filtered out, matching the existing "hidden surfaces stay queryable in memory; surface visibility is the user's call" contract Lifebooks already follow.
- New `applyLifebookFilter(entries, lifebookDomain)` is the load-bearing piece. Pure function — intersects the registry results with the selected Lifebook's `suggestedCapabilities` set. Three short-circuit cases documented in the function docstring: no filter selected, Lifebook not in cached list, Lifebook has empty `suggestedCapabilities` (extractor proposed nothing yet — show everything rather than collapse to zero).
- Empty-state copy surfaces the active Lifebook name (`No results found for the "Health" Lifebook.`) so the user can immediately tell *why* nothing's showing.

**Internal cleanup**

The three `renderRegistryResults(userId, q, category)` call sites collapsed to a single `renderRegistryResults(userId, readRegistryFilterState())` form via a new `readRegistryFilterState()` helper. Adding the Lifebook filter without this would have meant editing five separate call sites; the next filter to land only has to touch two places now (the dropdown markup and the state reader).

- **No DB migration.** The intersection set already lives on `lifebooks.suggested_capabilities`, populated by the domain-extraction worker that shipped in PR #242. Adding a `domain_name` column on `mcp_servers` would have been the heavier approach; this one stays purely on the read side.
- **No server-side filter.** The filter runs in JS after the API round-trip. The registry response is small (it's already pre-filtered by category) and lifebooks rarely have more than ~10 suggested capabilities each, so the client-side intersect is cheap.
- **No filter on the Installed section** (yet). Today the filter only narrows the *Browse Registry* results — that's where users discover new capabilities and most want to narrow by Lifebook. Filtering Installed is a follow-up if usage suggests it.

## Test plan

- [x] **Smoke-tested in Chrome** at `localhost:3202/#/capabilities` with mocked Lifebook + registry data:
  - **All Lifebooks** (initial): all 4 mocked entries shown.
  - **Health Lifebook**: filter narrows to `cap-fitbit` only (other Health-suggested entries aren't in the registry mock; the intersect handled it correctly).
  - **Money Lifebook**: filter narrows to `cap-mint` only.
  - **Back to All Lifebooks**: all 4 entries restored.
  - **Hidden Lifebook**: confirmed it doesn't appear in the dropdown options.
- [x] `node --check` clean on `capabilities.js`.

Smoke-test results captured live:

```json
{
  "dropdownPresent": true,
  "dropdownOptions": ["", "Health", "Money"],  // "Hidden" correctly absent
  "allEntriesCount": 4,
  "healthEntries": ["cap-fitbit"],
  "moneyEntries": ["cap-mint"],
  "allBackEntries": ["cap-fitbit", "cap-mint", "cap-github", "cap-spotify"]
}
```

## Notes for reviewers

- `applyLifebookFilter()` is intentionally pure (no DOM reads, no module-level state mutation) so it's portable to a vitest harness if `apps/web` ever gets one. The three short-circuit cases (no filter selected / Lifebook missing / empty suggested list) are documented inline.
- The "empty suggestedCapabilities means show everything, not collapse" call is the only judgment one here. The alternative ("extractor hasn't proposed anything yet, so this Lifebook filter should show zero") would have made the empty state harder to interpret — users would think the filter was broken. Worth flagging if you read it differently.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,55 @@
 All notable changes to SkyTwin will be documented in this file.
 
+## [unreleased] — Capabilities: filter the registry by Lifebook (#193 follow-up)
+
+Closes the "capability filter by domain on Capabilities page" item that
+PR #242 (#193 Child 1) explicitly deferred. The Capabilities page now
+carries a Lifebook dropdown alongside the existing category dropdown;
+selecting a Lifebook intersects the registry results with that
+Lifebook's `suggestedCapabilities: registryId[]` set so users browsing
+for a specific life domain see only the capabilities the domain
+extractor actually proposed for it.
+
+**No DB migration required.** The filter is purely client-side: the
+intersection set already lives on `lifebooks.suggested_capabilities`
+(populated by the domain-extraction worker that landed in #242). The
+capabilities page now fetches `/api/lifebooks/:userId` alongside its
+existing data and runs the intersect in `applyLifebookFilter()` — a
+pure helper that's trivial to lift to a vitest harness if/when one
+lands in `apps/web`.
+
+UX details:
+
+- Dropdown shows visible (non-hidden) Lifebooks only. Hidden Lifebooks
+  stay in memory but disappear from this dropdown, matching the
+  "hidden surfaces stay queryable in memory; surface visibility is the
+  user's call" contract Lifebooks already follow.
+- When the user has zero Lifebooks (domain extractor hasn't run, or
+  everything is hidden), the dropdown is omitted entirely — an empty
+  selector with no options would just confuse.
+- Empty-result state when a Lifebook filter narrows everything away
+  surfaces the active Lifebook name in the empty-state copy ("No
+  results found for the 'Health' Lifebook.") so the user can
+  immediately tell *why* nothing's showing.
+- A Lifebook with an empty `suggestedCapabilities: []` array means the
+  extractor proposed nothing yet — the filter shows everything rather
+  than collapsing to zero results, on the principle that "extractor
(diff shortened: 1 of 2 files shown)


jayzalowitz:
Round-2 review reply:

**Already fixed in round 1 (2c7eef2):**
- `applyLifebookFilter` now truly pure — `lifebooks` is the third argument with a cache-default at the call site.
- `data-action` removed from both `<select>` elements; the explicit `change` listeners are the sole entry point. Dead switch cases (`registry-filter-change`, `registry-category-change`) deleted.

**Round-2 finding fixed in eac5b7b:**
- CHANGELOG no longer describes `readRegistryFilterState()` as "pure" — it reads the DOM, so the description now calls it "isolated" instead.

PR #257, opened 2026-05-11
feat(#193 follow-up): provenance graph filter by Lifebook wing

jayzalowitz:
## Summary

Closes the **"provenance graph wing-filter consumer"** item that PR #242 (#193 Child 1) explicitly deferred. The lifebook page has been linking to `#/provenance?wing=<wingId>` since #242 shipped, but the provenance page couldn't honor that filter because nodes had no wing linkage. This PR adds the wing-id column + write-time population + API filter + frontend consumer end-to-end.

Two of three #193 follow-ups now closed (capabilities filter via PR #256; this one). One remaining: per-Lifebook briefing prose.

### Migration

**`packages/db/src/migrations/041-provenance-wing-id.sql`** — adds nullable `wing_id UUID` to `capability_provenance_nodes` plus a partial index `(user_id, wing_id, occurred_at DESC) WHERE wing_id IS NOT NULL`. Partial keeps the index small — only populated rows — while still serving per-wing hot-path queries. No FK to `lifebooks` (would block lifebook hard-delete); the frontend filter naturally excludes NULL rows.

### Write-time population

**`provenanceRepository.writeNode()`** auto-derives `wing_id` when the caller doesn't pass one explicitly. Logic: if the payload carries a `registryId`, look up the lifebook whose `suggested_capabilities` contains that id and stamp its `wing_id` on the node. Best-effort — a registryId not in any lifebook stays NULL. Explicit `wingId` argument always wins over the auto-derivation, so future call sites with their own wing context (e.g. an action invoked from inside a Lifebook page) can bypass the lookup.

### API

**`GET /api/capabilities/provenance-graph`** accepts a new optional `wing=<uuid>` query param. UUID-validated (returns 400 on bad shape, mirroring the existing `serverId` validation). Each node returned to the client includes `wingId: string | null`.

### Frontend

**`apps/web/public/js/pages/provenance-graph.js`** reads the `wing` param off the hash query string (the lifebook page's `#/provenance?wing=<uuid>` link), passes it through `fetchProvenanceGraph`, and renders a clearly-visible scoped-state indicator above the graph with a "Show all wings" button to clear the filter. UUID-validated client-side too so a malformed hash doesn't reach the API.

- **Backfill of `wing_id` for pre-migration provenance rows.** Older nodes stay NULL forever unless a future utility runs. The wing filter shows post-migration nodes; that's the right MVP for "Lifebook scoping" since the lifebook itself is a recent concept and the most relevant nodes are recent.
- **Per-domain briefing prose** — the third #193 deferred follow-up. Separate PR.

## Test plan

- [x] 3 new vitest cases in `provenance-graph-routes.test.ts`:
  - `wing` filter is forwarded as `wing_id` in the SQL params + WHERE clause.
  - Malformed `wing` UUID returns 400 with a `wing`-mentioning error.
  - Response nodes include `wingId` per node.
- [x] **Full api suite**: 547/547 passing (+3 new tests).
- [x] **Full workspace**: 70/70 turbo tasks green.
- [x] `pnpm build --concurrency=1` clean.
- [x] `node --check` clean on the JS file.

## Notes for reviewers

- The `resolveWingIdFromPayload` helper runs one extra SELECT per provenance-node write. On install/action node creation that's negligible (these aren't hot paths). For high-volume node writers (signal/entity), the payload typically doesn't have a `registryId`, so the helper short-circuits and never hits the DB.
- The partial index strategy means the per-wing query cost stays bounded even as the `capability_provenance_nodes` table grows with NULL-wing rows. CockroachDB supports partial indexes natively.
- Auto-derivation could disagree with explicit `wingId` if a caller passed one and a different lifebook later claimed the same registryId. The current contract — explicit wins — means the explicit value is authoritative. Worth flagging if you read it differently.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,65 @@
 All notable changes to SkyTwin will be documented in this file.
 
+## [unreleased] — Provenance graph: filter by Lifebook wing (#193 follow-up)
+
+Closes the **"provenance graph wing-filter consumer"** item that
+PR #242 (#193 Child 1) explicitly deferred. The lifebook page has
+been linking to `#/provenance?wing=<wingId>` since #242 shipped, but
+the provenance page couldn't honor that filter because nodes had no
+wing linkage. This PR adds the wing-id column + write-time
+population + API filter + frontend consumer end-to-end.
+
+### Migration
+
+- `041-provenance-wing-id.sql` — adds nullable `wing_id UUID`
+  column to `capability_provenance_nodes` plus a partial index
+  `(user_id, wing_id, occurred_at DESC) WHERE wing_id IS NOT NULL`
+  so per-wing graph queries are indexed without bloating the index
+  with the long tail of NULL rows. No FK to lifebooks (would block
+  lifebook hard-delete); the frontend filter naturally excludes
+  NULL rows.
+
+### Write-time population
+
+- `provenanceRepository.writeNode()` auto-derives `wing_id` when
+  the caller doesn't pass one explicitly: if the payload carries
+  a `registryId`, look up the lifebook whose
+  `suggested_capabilities` contains that id and stamp its
+  `wing_id` on the node. Best-effort — a registryId not in any
+  lifebook stays NULL. Explicit `wingId` argument always wins
+  over auto-derivation, so future call sites with their own wing
+  context can bypass the lookup.
+- Older rows written before migration 041 stay NULL forever
+  unless a future backfill utility runs.
+
+### API
+
+- `GET /api/capabilities/provenance-graph` now accepts an optional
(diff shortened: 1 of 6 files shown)


jayzalowitz:
Round-2 review reply: both findings addressed in 048ec85.

- **`pg-clear-wing-filter` re-render claim**: comment now describes both branches accurately. The hash-actually-changed branch DOES rely on the SPA router's hashchange path; the unusual already-on-base-hash branch falls through to an explicit `renderProvenanceGraph` call.
- **`resolveWingIdFromPayload` docstring**: rewrote the "returns null when" list. The signature already excludes "not an object" at compile time, so the runtime checks are payload=undefined or registryId missing/non-string.

PR #275, opened 2026-05-13
feat(#251 Phase 2): relationshipTier — second retrieval axis

jayzalowitz:
## Summary

Phase 2 of the multi-phase #251 plan. Adds a second retrieval dimension on top of Phase 1's additive bonuses: **`relationshipTier`**, derived from bidirectional thread count over the last 90 days. Composes additively with `authoringTier`.

Stacked on top of #274 (Phase 1) — but Phase 1's content is already in main via the squash-merge, so this PR's diff against main shows only the Phase 2 additions.

## Tier bands

| Tier | Threshold | Examples |
|---|---|---|
| `core` | ≥ 10 bidirectional days in 90d | Spouse, manager, daily collaborators |
| `frequent` | 3–9 | Team members, vendors actually engaged |
| `occasional` | 1–2 | Sporadic |
| `stranger` | 0 | Cold senders, broadcast, never-replied |

## Bonuses (normal calibration, additive, promote-only)

Same scale as Phase 1's authoring bonuses; same lesson — `stranger`/0 instead of demote:

| Band | Bonus |
|---|---|
| `core` | +0.004 |
| `frequent` | +0.002 |
| `occasional` | 0 |
| `stranger` | 0 |

Combined effects:

- `inbox_personal` + `core` = +0.004 → a question from your manager outranks a same-topic newsletter
- `user_sent_originated` + `core` = +0.005 + +0.004 = **+0.009** → decisive on ambiguous queries
- `user_sent_originated` + `stranger` = +0.005 → authored from a stranger still gets the authored boost; no rel demote

## Engine

- **`tier-weights.ts`** gets `RelationshipTier` enum, `REL_SPARSE` / `REL_NORMAL` / `REL_DENSE` tables, `relationshipTierFromThreadCount` helper. `tierBonus` composes the relationship dimension additively after the authoring dimension.
- **`computeBidirectionalThreadCounts(userId, windowDays)`** (new adapter helper): SQL CTE that splits user's gmail signals into `received` (no SENT label) vs `sent` (has SENT label) and joins by contact address within the 90d window. Returns `Map<contactAddress, bidirectionalDays>`.
- **In-memory mirror** matches the CRDB semantics bit-for-bit.

## Worker

- **`runRelationshipTierBackfillJob(userId)`** (new): pulls the count map, walks recent pages, computes the tier band per page from `metadata.fromAddress`, writes via `updatePageMetadata` only when the tier differs. Idempotent.
- Scheduled daily in `apps/worker/src/index.ts` (`RELATIONSHIP_TIER_BACKFILL_INTERVAL_MS = 24h`). Per-user, failures isolated.

## Dashboard

- Recent pages indexed table now has a **Relationship** column with color-coded badges (`core` = green, `frequent`/`occasional` = neutral, `stranger` = muted) alongside the authoring-tier badge.
- `/api/memory-config/dashboard` payload's `pages.recent[]` includes `relationshipTier` projected from metadata.

## Tests

- 8 new unit tests in `tier-weights.test.ts`: per-tier bonuses, composition with authoring, sparse/dense calibration scaling, pinned-override composition, hidden-override drop, threshold helper.
- 7 new worker tests in `relationship-tier-backfill.test.ts`: per-tier derivation, lower-case normalization, unchanged detection, skipped (no `fromAddress`), failure isolation, find-throws → empty summary, 0-affected → counted as failed.

## Test plan

- [x] `pnpm build --concurrency=1` → 35/35 packages
- [x] `pnpm test` → 70/70 turbo tasks green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,51 @@
 All notable changes to SkyTwin will be documented in this file.
 
+## [unreleased] — relationshipTier: second retrieval axis (#251 Phase 2)
+
+Adds a second dimension to gbrain's tier weighting: how strong the user's back-and-forth has been with the contact over the last 90 days. Composes additively with `authoringTier` from Phase 1.1.
+
+### Tier bands
+
+Computed from bidirectional thread count (days where the user both sent to AND received from the contact) over the last 90 days:
+
+| Tier | Threshold | Examples |
+|---|---|---|
+| `core` | ≥ 10 bidirectional days | Spouse, manager, daily collaborators |
+| `frequent` | 3–9 | Team members, vendors actually engaged with |
+| `occasional` | 1–2 | Sporadic exchanges |
+| `stranger` | 0 | Cold senders, broadcast lists, never-replied |
+
+### Bonuses (normal calibration, additive, same scale as authoring)
+
+- `core`: +0.004 — a page from a close contact gets a small boost
+- `frequent`: +0.002
+- `occasional`: 0
+- `stranger`: 0 (promote-only, matching Phase 1.1's lesson)
+
+Combined effect: an `inbox_personal` page from a `core` contact gets +0.004 (climbs ahead of a same-topic newsletter without authored sibling); a `user_sent_originated` to a `core` contact gets +0.005 + +0.004 = +0.009 (decisive on ambiguous queries).
+
+### Engine
+
+- **`tier-weights.ts`** gets `RelationshipTier` enum, `REL_*` calibration tables, `relationshipTierFromThreadCount` helper, and `tierBonus` now composes the relationship dimension additively after the authoring dimension.
+- **`computeBidirectionalThreadCounts(userId, windowDays)`** new adapter helper. SQL: JOIN `brain_signals` self-join on contact address within the window, restricted to rows where one side has the `SENT` label and the other doesn't. Returns `Map<contactAddress, bidirectionalDays>`.
+- **In-memory mirror** matches the CRDB SQL behaviour bit-for-bit.
+
+### Worker
+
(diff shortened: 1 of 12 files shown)


jayzalowitz:
Addressed Copilot finding on bidirectional thread query in 1087f84:

1. **Angle-bracket extraction** — switched `TRIM(BOTH '<>' FROM …)` to `regexp_replace(…, '.*<([^>]+)>.*', '\1')` so `Display Name <addr@x.com>` correctly yields `addr@x.com`. Mirrors `extractBareAddress` in `packages/connectors/src/authoring-tier.ts`.
2. **to/cc list splitting** — switched to `LATERAL unnest(string_to_array(COALESCE(to,'') || ',' || COALESCE(cc,''), ','))` so multi-recipient sent emails produce per-contact rows. Was previously treating the whole list as one address.
3. **In-memory mirror also updated** to include cc (was to-only), keeping both backends identical.

7 new tests in `in-memory-repository.test.ts > computeBidirectionalThreadCounts` pin the parsing: display-name extraction, bare-address pass-through, to-list splitting, cc inclusion, received-only contacts (no bidirectional), distinct-day counting, windowDays cutoff.

jayzalowitz:
Follow-up to the SQL fix: Copilot's second-pass review on PR #277 caught another real bug in the same query. `data->'labels' @> '"SENT"'::JSONB` returns NULL when `labels` is missing — and NULL is falsy in WHERE — so any signal without an explicit `labels` field was silently dropped from BOTH the received AND sent CTEs (the received side's `NOT (NULL)` is also NULL/falsy).

Fixed in 965d243 by wrapping both predicates in `COALESCE(..., false)` so missing `labels` falls into the received bucket (matching the in-memory mirror's `Array.isArray(labels) ? labels : []` behaviour). New test `treats missing labels as received (not silently dropped)` pins it.

PR #276, opened 2026-05-13
feat(#251 Phase 3): cross-channel tier — calendar events get authoringTier

jayzalowitz:
## Summary

Phase 3 of #251 — extends the `authoringTier` vocabulary to calendar events. Calendar entries now flow through the same RRF-fold + Phase-1 bonuses + Phase-2 relationship composition as Gmail signals.

**Stacked on Phase 2 (#275).** Will rebase clean once Phase 2 merges.

## Mapping

| Calendar shape | AuthoringTier |
|---|---|
| User organized the event (organizer email = `self`) | `user_sent_originated` |
| Auto-generated event — Google Contacts birthdays, holidays feed, weather | `inbox_automated` |
| Multi-attendee invite (> 2 attendees) | `inbox_broadcast` |
| 1-on-1 invite or shared calendar entry | `inbox_personal` |

`user_sent_reply` and `inbox_newsletter` aren't applicable to calendar and aren't produced.

## Why reuse the vocabulary instead of extending it

The #251 spec deliberately named tier values to generalize across channels. `user_sent_originated` reads as "user authored, fresh" not "email the user sent." Reusing the same enum keeps everything downstream channel-agnostic: one bonus table, one set of weights in `tier-weights.ts`, one eval covers both Gmail and calendar.

## Engine

- New `classifyCalendarAuthoringTier(input)` in `packages/connectors/src/calendar-authoring-tier.ts`. Recognizes the Google Contacts birthdays feed (`addressbook#contacts@`), holiday calendars (`en.usa#holiday@`), and weather feeds via regex.
- `GoogleCalendarConnector.eventToSignal` calls the classifier and stamps both `authoringTier` and `from` (organizer email) on `signal.data`. The `from` field matters because `EmbeddedGbrainMemoryPort.buildPageMetadata` reads it to populate `metadata.fromAddress` — which is what Phase 2's relationship-tier worker queries and what the per-sender bulk-hide UI from PR #270 sends to.

## Tests

- 8 cases in `calendar-authoring-tier.test.ts` — user-organized (case-insensitive), birthdays/holidays/weather → automated, multi-attendee → broadcast, 1-on-1 → personal, edge cases for empty `selfEmail`.

## Out of scope

- GitHub authoring tier — no GitHub connector exists in this codebase. The vocabulary is already channel-agnostic, so a future GitHub connector can stamp tiers the same way without any engine changes.

## Test plan

- [x] `pnpm build --concurrency=1` → 35/35 packages
[...]

PR #277, opened 2026-05-13
feat(#251 Phase 4): draft_email candidate generator

jayzalowitz:
## Summary

Phase 4 of #251 — the marquee feature Layers 1+2+3 were building toward. Adds `DraftEmailCandidateGenerator` that drafts a reply body in the user's voice using their authored corpus as few-shot grounding.

This builds on the tier-aware retrieval shipped in Phase 1.1 (additive bonuses), Phase 2 (relationshipTier axis), and Phase 3 (cross-channel tiers). When the memory layer returns the top-K most similar emails the user has written, those become the voice/style training for the LLM.

## What ships

- **`DraftEmailCandidateGenerator`** (`packages/decision-engine/src/strategies/draft-email-candidate.ts`) — implements `CandidateGenerator`. Fires when `decision.domain === 'email'` and `rawData.requiresResponse` is truthy. Produces one `CandidateAction` with `actionType: 'draft_email'`.
- **`AuthoredExamplesPort`** interface — minimal port the generator needs from memory. Decoupled from `@skytwin/memory-port` so the decision-engine package stays leaf-level. Implementations live wherever the caller wires things up (typically `MemoryPort.searchAuthored` with a tier filter).
- **`buildDraftPrompt(input)`** — exported helper that renders the few-shot prompt. Structure is load-bearing for voice match, so it's testable directly.

## Wiring approach: opt-in

Exported from the package but **NOT wired into `DecisionMaker.evaluate` by default**. The deploy decision (which LLM client, when to flip it on, eval gates, cost gating) lives outside the decision-engine layer. Callers instantiate it explicitly. This phase ships the building block — the eval / rollout work follows.

## Safety

- `reversible: true` — the candidate itself is reversible; the user's send is what crosses the threshold (policy engine handles that)
- `MODERATE` confidence when ≥ 3 authored examples were available, `LOW` otherwise
- Always requires explicit approval at v1 regardless of trust tier (no auto-send)
- Fail-open on memory error (drafts without grounding, warns about voice match in reasoning)
- Fail-closed on LLM error (returns no candidate rather than shipping a template-y draft)
- `parameters.examplesUsed` surfaces grounding strength to the approval UI

## Prompt shape

- System prompt: match voice/length/opening/closing/vocabulary from few-shot examples; output ONLY the body, no subject/signature/preamble; 1–4 short paragraphs
- Few-shot examples truncated at `MAX_EXAMPLE_CHARS = 800` (stops a 5KB email from dominating context)
- Default `DEFAULT_AUTHORED_EXAMPLE_COUNT = 6` (~1500 prompt tokens for examples)
- Memory query composed from inbound subject + first body line + from address, capped at 500 chars

## Test plan

- [x] 20 unit tests in `__tests__/draft-email-candidate.test.ts` covering: domain gating (email + requiresResponse), candidate shape and parameter projection, confidence bands (LOW < 3 examples, MODERATE ≥ 3), fail-open / fail-closed on subsystem failure, rawData fallbacks (snippet→body, messageId→emailId), `buildDraftPrompt` rendering (with/without examples, truncation, placeholder copy), query construction (500-char cap, empty-field stripping)
- [x] Full decision-engine suite: 129 tests green
- [x] Full workspace build (serial): 35 packages
- [x] Full workspace test: 70 tasks pass

## Stacking note

Stacked atop Phase 2 (#275) and Phase 3 (#276); base is `main` but the diff includes Phase 2+3 commits until those merge. Once Phase 2/3 land, this PR rebases cleanly to its own delta (one commit, ~700 LOC).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,38 @@
 All notable changes to SkyTwin will be documented in this file.
 
+## [unreleased] — draft_email candidate generator (#251 Phase 4)
+
+The marquee feature Layers 1+2+3 were building toward. When an inbound email needs a reply, a new candidate generator drafts the body in the user's voice using their authored corpus as few-shot grounding.
+
+### What's new
+
+- `packages/decision-engine/src/strategies/draft-email-candidate.ts` — `DraftEmailCandidateGenerator` implementing `CandidateGenerator`. Fires only when `decision.domain === 'email'` and `rawData.requiresResponse` is truthy. Produces one `CandidateAction` with `actionType: 'draft_email'`.
+- `AuthoredExamplesPort` interface — the minimal port the generator needs from the memory layer. Decoupled from `@skytwin/memory-port` so the decision-engine package stays leaf-level.
+- `buildDraftPrompt(input)` — exported helper that renders the few-shot prompt. Structure is load-bearing for voice match, so it's testable directly.
+
+### Wiring approach: opt-in
+
+The generator is **exported but not wired into `DecisionMaker.evaluate` by default**. The deploy decision (which LLM client, when to flip it on, eval gates) lives outside the decision-engine layer. Callers instantiate it explicitly and add it to their generator list. This phase ships the building block.
+
+### Confidence and safety
+
+- `reversible: true` (the candidate is reversible up to the user clicking Send; the policy engine handles the actual send threshold)
+- `MODERATE` confidence when ≥ 3 authored examples were available, `LOW` otherwise — the approval UI surfaces "drafted from N of your prior emails" via `parameters.examplesUsed` so the user sees the grounding strength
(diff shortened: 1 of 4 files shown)


jayzalowitz:
Addressed both findings in bcfea28:

1. **NULL-labels silently dropped** — real bug. `data->'labels' @> '"SENT"'::JSONB` returns NULL when labels is missing, which is falsy in WHERE; `NOT (NULL)` is also NULL/falsy → both received and sent CTEs silently dropped any signal without explicit `labels`. Wrapped in `COALESCE(..., false)` on both sides. The in-memory mirror already treated missing labels as `[]` so the two backends were silently divergent. New test pins it.
2. **temperature typo** — fixed.

The other finding (to-list / angle-bracket parsing) is the same as Phase 2's, already addressed via cherry-pick of a51cd09.

PR #278, opened 2026-05-13
deps: upgrade express + @types/express to v5 (carries forward #105 + type fix)

jayzalowitz:
## Summary

Closes #105. Same `express` 4 → 5 + `@types/express` 4 → 5 bump Dependabot proposed, plus the 2-line type-narrowing fix needed to keep `apps/api/src/routes/sessions.ts` compiling under @types/express@5.

## Why a new PR

GitHub Actions security policy stops the PR-level workflows from re-running on a Dependabot branch after a non-bot push — even after close/reopen. Carrying the same commits over to a maintainer branch unblocks CI.

## What changed

- `apps/api/package.json`, `apps/twin-mcp-server/package.json`, `apps/web/package.json`: `express ^4.18.2 → ^5.2.1`, `@types/express ^4.17.21 → ^5.0.6`
- `pnpm-lock.yaml`: regenerated by Dependabot
- `apps/api/src/routes/sessions.ts:67` and `:96`: existing `if (!userId)` / `if (!sessionId)` guards now also check `typeof === 'string'` — @types/express@5 widens destructured `req.params` fields to `string | string[] | undefined` because routing patterns with splat (`*`) can produce arrays, so the existing truthy check alone no longer narrowed to `string`

## What's notably NOT changed

Every other `const { X } = req.params` in `apps/api/` and `apps/twin-mcp-server/` continues to compile, because TS only flagged the two call sites that pass the destructured value straight into a strictly-typed repository method. If those other sites *do* break in practice, the same `typeof === 'string'` pattern applies and is a quick follow-up.

## Test plan

- [x] Type narrowing fix is local to sessions.ts; no behavior change beyond returning 400 instead of 500 if someone constructs a request that hits the route with an array-shaped param (which shouldn't be reachable through normal routing)
- [ ] CI Test + eval suite green on this branch
- [ ] Smoke-test sessions list/revoke endpoints once a build artifact is available

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -38,11 +38,11 @@
     "@skytwin/shared-types": "workspace:*",
     "@skytwin/twin-model": "workspace:*",
     "bonjour-service": "^1.3.0",
-    "express": "^4.18.2",
+    "express": "^5.2.1",
     "tweetnacl": "^1.0.3"
   },
   "devDependencies": {
-    "@types/express": "^4.17.21",
+    "@types/express": "^5.0.6",
     "tsx": "^4.7.0",
     "typescript": "^6.0.3",
     "vitest": "^1.3.0"
--- a/apps/api/src/routes/sessions.ts
+++ b/apps/api/src/routes/sessions.ts
@@ -65,7 +65,7 @@ export function createSessionsRouter(): Router {
     try {
       const { userId } = req.params;
       if (typeof userId !== 'string' || !userId) {
-        res.status(400).json({ error: 'Missing userId' });
+        res.status(400).json({ error: 'Missing or invalid userId' });
         return;
       }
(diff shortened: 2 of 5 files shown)


jayzalowitz:
Addressed Copilot review:

1. **Error message specificity** (line 69, 96) — fixed in 0c02756. Both guards now return "Missing or invalid userId" / "Missing or invalid sessionId" so clients see an accurate 400 when the param is array-shaped, not absent.

2. **Route-level tests for /api/sessions** (line 68, 92) — deferring as a follow-up. The existing guard behavior is unchanged for the request shapes that real clients send (path params come through as strings); the new typeof check is defense-in-depth against the type widening, and adding the testing surface for sessions routes is a larger scope-of-work that doesn't belong on this PR. I'll open a follow-up issue.

Issue #371, opened 2026-05-25
P1.1 RiskAssessment fork in auto-execute path


jayzalowitz:
**Parent:** #361

## Context

SkyTwin's whole safety story rests on Safety Invariant #7: every `CandidateAction` carries a `RiskAssessment`, and that assessment is what the execution router uses to decide which adapter is allowed to run the action and with what guardrails. The decision-maker produces one — six dimensions, an `overallTier`, reasoning, an `assessedAt` timestamp — and persists it to `candidate_actions.risk_assessment`. The execution router refuses to run anything without one (CLAUDE.md § Safety Invariants, item 7).

Today the assessment the decision-maker produced is **discarded** at both execution boundaries. `apps/api/src/routes/events.ts` (auto-execute path) re-derives a fresh assessment from the explanation's coarse `riskTier` enum. `apps/api/src/routes/approvals.ts` (human-approved path) constructs a synthetic assessment with hardcoded `LOW` on every dimension. The router-level `actionId` backstop catches the case where the assessment is for the wrong action, but it cannot catch "this assessment claims LOW across every dimension when the decision-maker actually computed HIGH on financial impact."

In product terms: a candidate the policy engine should have escalated to confirmation can slip through with the wrong adapter and the wrong risk modifier applied. For a launch-day stranger trusting the twin with their inbox or their wallet, this is the single category of bug they cannot personally detect or recover from.

## Root Cause Analysis

Three structural decisions compound into the fork:

1. **The persisted assessment lives in the database; the in-memory `DecisionOutcome` carries only the **selected** one.** `packages/decision-engine/src/decision-maker.ts:186-191` calls `assessRisk()` per candidate, stores them in a `Map<string, RiskAssessment>`, persists each via `decisionRepository.saveRiskAssessment(assessment)`, and only attaches the **selected** candidate's assessment to `outcome.riskAssessment` (line 263). When the API route consumes `outcome` after a round-trip through persistence, it has the selected assessment available but the call sites never read it.

[...]

PR #424, opened 2026-05-26
P1.12 #412: DecisionOutcome carries per-candidate RiskAssessments

### LINKED ISSUE #412: P1.12 DecisionOutcome carries only selected candidate's RiskAssessment (divergence risk)
**Parent:** #361

## Context

While verifying #371 (RiskAssessment fork in auto-execute path), the safety review surfaced a related but distinct hazard: the `DecisionOutcome` object that is persisted and read by downstream code carries **only the selected candidate's** RiskAssessment, even though the DB schema persists the assessment for **every** candidate. If anything in the system ever re-selects a candidate post-decision (a user picks a different candidate from the queue, a manual override, a future "swap candidate" UX, or a bug), the `outcome.riskAssessment` no longer matches `outcome.selectedAction`.

The downstream consumer most affected is the execution-router invariant check at `execution-router.ts:65-83`, which compares `action.id !== riskAssessment.actionId` as a safety backstop. If the outcome were ever rebuilt with a different selectedAction but the old riskAssessment, the backstop would catch it — but only because the IDs would mismatch. The safer pattern is to either (a) carry **all** candidates' assessments on the outcome so the right one is always retrievable by ID, or (b) re-fetch the assessment by action ID at every consumer boundary.

This is structurally the same class of bug as #371: a single source of truth in the decision-maker, then a fork at the outcome boundary that risks divergence. #371 fixes the auto-execute path; this issue closes the persistence-layer drift.

## Root Cause Analysis

`packages/decision-engine/src/decision-maker.ts:258-274` builds the `DecisionOutcome` with `allCandidates: candidates` (plural — every candidate) but `riskAssessment: selectedAssessment` (singular — only the picked one). This was a reasonable shortcut when the outcome was treated as immutable, but it bakes in an assumption: "the only assessment that matters is the one for the candidate I selected." Three drift paths break that assumption:

1. **Manual re-selection from the approval queue.** A user looking at a candidate's queue entry could in principle approve a *different* candidate than the one the engine selected. The current API doesn't support this, but the DB schema does — `candidate_actions` has rows for every candidate.
2. **Cross-process reads.** The worker, the briefing job, and the audit-trail API all read decision outcomes. If any of them re-derives the "best" candidate from `allCandidates`, the `riskAssessment` field is misleading.
3. **Future "second-opinion" features.** Any post-hoc UI that lets a user explore "what would have happened if I'd picked candidate 2" needs candidate 2's assessment. Today it's stored in the DB but not in the outcome — so the UI has to do an extra round-trip and stay in sync.

The DB schema is already doing the right thing: `candidate_actions.risk_assessment` JSONB column persists every candidate's full assessment. The bug is only in the in-memory outcome object.

## Verified Current State

Verified 2026-05-25 against `jayzalowitz/launch-skytwin-demo @ 8bb5379`.

| File | Lines | Behavior |
|---|---|---|
| `packages/decision-engine/src/decision-maker.ts` | 258-274 | Builds `DecisionOutcome` with `allCandidates: candidates` (all), `riskAssessment: selectedAssessment` (only selected). |
| `packages/decision-engine/src/decision-maker.ts` | 186-191 | `selectedAssessment` is the assessment of the picked candidate; assigned inside the per-candidate loop. |
| `packages/db/src/repositories/decision-repository.ts` | 218-225 | INSERT into `candidate_actions` persists `risk_assessment` JSONB per candidate — full coverage. |
| `packages/db/src/repositories/decision-repository-adapter.ts` | 268-275 | `getRiskAssessment(candidateActionId)` exists and reads the per-candidate column. Currently called by no production code. |
| `packages/execution-router/src/execution-router.ts` | 65-83 | Backstop compares `action.id !== riskAssessment.actionId`. Catches ID mismatch, not "assessment is stale relative to a re-selected action." |
| `apps/api/src/routes/events.ts` | 558-574 | (Per #371) Constructs a fresh `RiskAssessment` rather than reading from outcome — see #371. |
| `apps/api/src/routes/approvals.ts` | 419-434 | (Per #371) Constructs hardcoded-LOW assessment — see #371. |

### Audit Table — Every place RiskAssessment is created, persisted, or read

| Code site | Role | Reads / Writes | Risk |
|---|---|---|---|
| `decision-maker.ts:186-191` | Source of truth | Writes per-candidate | Correct |
| `decision-maker.ts:263` | Outcome assembly | Writes only selected to outcome | **Divergence point** |
| `decision-repository.ts:218-225` | Persistence | Writes per-candidate to DB | Correct |
| `decision-repository-adapter.ts:268-275` | Persistence read | Has `getRiskAssessment(id)` | Unused — should be |
| `events.ts:558-574` | Auto-execute path | Constructs new (#371) | **#371 fixes this** |
| `approvals.ts:419-434` | Approval execute path | Constructs new with LOW (#371) | **#371 fixes this** |
| `execution-router.ts:65-83` | Backstop | Reads from passed assessment | Correct (validates ID match) |

The two columns to align: #371 makes `events.ts` + `approvals.ts` read `outcome.riskAssessment`. This issue then ensures `outcome.riskAssessment` doesn't drift from `outcome.selectedAction`.

## What's Working Well (Do Not Touch)

- DB persistence is correct — `candidate_actions.risk_assessment` JSONB per row.
- `execution-router.ts:65-83` ID-match backstop is correct as a defense-in-depth check. Don't remove it after this issue lands.
- `decision-repository-adapter.getRiskAssessment(candidateActionId)` exists and works — this issue starts using it, doesn't rewrite it.
- The decision-maker's per-candidate loop at `186-191` is the right source-of-truth pattern.

## Proposed Change

Two coordinated changes:

### Change 1 — Outcome carries all candidates' assessments

`DecisionOutcome` gains a new optional field `allRiskAssessments: RiskAssessment[]` (or `Map<string, RiskAssessment>` keyed by `candidateActionId`). The existing `riskAssessment: RiskAssessment` field stays for backwards-compatibility but is documented as "convenience pointer to the selected candidate's assessment — for the source of truth use allRiskAssessments[].find(a => a.actionId === selectedAction.id)".

### Change 2 — Consumers read by ID, not by trust

Any code path that needs a specific candidate's risk should:
1. Take the actionId.
2. Find it in `allRiskAssessments` (in-memory) or call `getRiskAssessment(actionId)` (DB).
3. Never assume `outcome.riskAssessment` is for the action they care about.

In practice for this issue's scope, the only consumer that matters today is the execution-router boundary (already does ID-match via the backstop). The cleanup is making sure no future consumer accidentally reaches for `outcome.riskAssessment` without an ID check.

### Type change

```ts
// packages/shared-types/src/decision.ts
export interface DecisionOutcome {
  id: string;
  decisionId: string;
  selectedAction: CandidateAction | null;
  allCandidates: CandidateAction[];
  /** @deprecated Use allRiskAssessments. Kept as a convenience pointer; equals the assessment whose actionId matches selectedAction.id. */
  riskAssessment: RiskAssessment | null;
  /** New: every candidate's assessment, keyed by actionId. Source of truth for the in-memory outcome. */
  allRiskAssessments: RiskAssessment[];
  // ... rest unchanged
}
```

### Decision-maker update

`decision-maker.ts:258-274` collects assessments into a local array as it loops candidates (already has the data; just don't throw it away). Sets both `riskAssessment` (selected) and `allRiskAssessments` (all).

### Helper

```ts
// packages/decision-engine/src/decision-helpers.ts (new or existing)
export function getAssessmentForAction(
  outcome: DecisionOutcome,
  actionId: string,
): RiskAssessment | null {
  return outcome.allRiskAssessments.find(a => a.actionId === actionId) ?? null;
}
```

## Acceptance Criteria

1. `DecisionOutcome` type has a new `allRiskAssessments: RiskAssessment[]` field; the existing `riskAssessment` field is marked `@deprecated` but retained.
2. `decision-maker.ts:258-274` populates both fields. `allRiskAssessments` has exactly one entry per candidate; every entry's `actionId` matches a candidate in `allCandidates`.
3. `getAssessmentForAction(outcome, actionId)` helper exists in `packages/decision-engine/`.
4. Unit test: a decision with N candidates produces an outcome with N entries in `allRiskAssessments`, all with matching actionIds.
5. Unit test: `getAssessmentForAction` returns null for a missing actionId.
6. After #371 lands, both `events.ts` and `approvals.ts` use `getAssessmentForAction(outcome, candidate.id)` rather than `outcome.riskAssessment` directly. (Verify by grepping; the deprecation comment is the canary.)
7. No regression: existing decision-pipeline tests continue to pass with both fields populated.

## Testing Plan

| Layer | What | Count |
|---|---|---|
| Unit | `decision-maker` produces N assessments for N candidates | +1 |
| Unit | `getAssessmentForAction` returns the right one by ID | +1 |
| Unit | `getAssessmentForAction` returns null for missing ID | +1 |
| Integration | Save → load decision; assert `allRiskAssessments` round-trips via the existing per-candidate DB rows | +1 |

Note: the persistence layer already writes per-candidate to `candidate_actions.risk_assessment`. The integration test verifies the read path reassembles them into `allRiskAssessments`.

## Rollback Plan

The new `allRiskAssessments` field is additive. Old consumers reading `riskAssessment` keep working. Rollback = revert the PR. No schema migration (DB already stores per-candidate).

## Effort Estimate

| Component | Hours |
|---|---|
| Type addition in `shared-types` + dist rebuild | 0.5h |
| Decision-maker outcome assembly update | 0.5h |
| Helper + tests | 1h |
| Persistence read path: reassemble allRiskAssessments | 0.5h |
| Coordinate with #371 to switch consumer call sites | 0.5h |

**Total: ~3h.** Land after or in the same PR as #371.

## Files Reference

| File | Change |
|---|---|
| `packages/shared-types/src/decision.ts` | Add `allRiskAssessments` field; mark `riskAssessment` `@deprecated` |
| `packages/decision-engine/src/decision-maker.ts:258-274` | Populate both fields |
| `packages/decision-engine/src/decision-helpers.ts` (new) | `getAssessmentForAction` helper |
| `packages/db/src/repositories/decision-repository.ts` | Read path: hydrate `allRiskAssessments` from `candidate_actions` rows |
| `packages/decision-engine/__tests__/decision-maker.test.ts` | New tests |

## Out of Scope

- Removing the deprecated `riskAssessment` field. Defer to a later cleanup; backwards-compat for one launch cycle.
- The auto-execute path's wrong-assessment bug — that's #371's job.
- The execution-router backstop change — keep the ID-match check, it's defense in depth.
- "What if a user picks a different candidate from the queue" UX. Not yet a feature.

## Related

- #361 Epic D (parent)
- #371 P1.1 — fixes the auto-execute path's assessment fork; this issue closes the parallel divergence at the outcome-assembly boundary
- CLAUDE.md Safety Invariant #7 — risk assessment mandatory; this issue strengthens "mandatory AND correct"

jayzalowitz:
## Summary

Closes #412. Same divergence-hazard class as #371, at the in-memory persistence-shape layer.

Pre-fix: `DecisionOutcome.riskAssessment` was the **selected** candidate's assessment only. The DB persisted per-candidate (`candidate_actions.risk_assessment` JSONB) but the in-memory outcome lost that completeness — any code path that re-selected post-decision (manual override, future "swap candidate" UX, a bug) left `riskAssessment` stale relative to `selectedAction`. Execution-router actionId-match backstop catches the mismatch, but only after the wrong-tier adapter is picked.

## Changes

- `packages/shared-types/src/decision.ts`: `DecisionOutcome` gains optional `allRiskAssessments: RiskAssessment[]`. Optional for backward compat (pre-#412 test fixtures + proactive-evaluator synthetic outcomes don't carry it). `riskAssessment` re-documented as a convenience pointer to the selected candidate.
- `packages/decision-engine/src/decision-maker.ts`: populates `allRiskAssessments` from the per-candidate `assessments` Map step 4 already builds, in the same iteration order as `allCandidates`.
- `packages/decision-engine/src/decision-helpers.ts` (new): `getAssessmentForAction(outcome, actionId)` prefers the snapshot, falls back to the legacy field only when the queried id matches the selected one, returns null on miss → caller should fall back to `decisionRepositoryAdapter.getRiskAssessment(actionId)`.
- `packages/decision-engine/src/proactive-evaluator.ts`: explicitly sets `allRiskAssessments: []` on synthetic outcomes from `whatWouldIDo`.

## Tests

- 6 new tests in `decision-helpers.test.ts` cover the matrix: id-match, miss, legacy fallback (id matches selected vs doesn't), empty array, prefer-new-over-legacy when both populated.
- **207 decision-engine tests pass (+6); full monorepo green** (713 api, 177 policy-engine, 63 mempalace, etc.).

## Verified

- `pnpm build --concurrency=1` ✓
- `pnpm test` ✓

## Compatibility

No production consumers reach for `outcome.riskAssessment` with a non-selected id today (verified via grep). This is defense-in-depth: any future consumer of a specific candidate's risk goes through `getAssessmentForAction` instead of assuming the convenience pointer is for the candidate they care about. The execution-router's `action.id !== riskAssessment.actionId` backstop is unchanged and remains the safety net.


jayzalowitz:
Addressed all 4 Copilot findings:

| # | Finding | Resolution |
|---|---------|------------|
| 1 | Unused `ConfidenceLevel` import → `noUnusedLocals` build break | Removed in 0983393 |
| 2 | `allRiskAssessments` + `allCandidates` correlation claim falsified by `.filter(undefined)` | Reworded comment in decision-maker.ts — deterministic walk + explicit direction to use the id-keyed `getAssessmentForAction` helper |
| 3 | `riskAssessment` doc overstates equality (selectedAction can be null) | Qualified with the 3 conditions for equality + explicit null otherwise |
| 4 | proactive-evaluator comment omits legacy fallback | Rewrote to note both fields intentionally null on synthetic outcomes, helper gives uniform null path |

All 3 in 0f35c9b are documentation accuracy fixes — no behavior change. 207 decision-engine tests still green.

PR #422, opened 2026-05-26
P1.7 #377: user-facing surface for OAuth re-auth (stacked on P1.9)

jayzalowitz:
**Stacked on #421 (P1.9 kill switch).** Will rebase against main once #421 lands.

## Summary

Closes #377. Pre-fix: when Google revoked SkyTwin's refresh token, the worker correctly stopped polling — but the dashboard rendered "Listening" and the user noticed days later. This PR adds the single missing user-facing signal.

### DB
- Migration `060-connector-health.sql`: `connector_health(user_id, connector_name, status, error_code, last_success_at, last_failure_at, updated_at)`, PRIMARY KEY (user_id, connector_name), ON DELETE CASCADE.
- New `connectorHealthRepository.upsert` + `findByUser`. Re-exported from `@skytwin/db`.

### Worker
- `apps/worker/src/index.ts:pollUserConnectors` upserts `status='needs_reauth'` on the `OAuthRefreshError.permanent` branch (alongside the existing circuit-breaker force-trip).
- Per-connector success upserts `status='connected'` keyed on a new `thisConnectorFailed` flag (not loop-wide `hadFailure`) so a working Calendar isn't blocked by a failing Gmail.
- DB writes wrapped in try/catch — must not break the circuit-breaker logic.
- New `extractErrorCode()` helper parses `invalid_grant` / `unauthorized_client` / etc. from the error message so the banner can render conditional copy.

### API
- New `apps/api/src/routes/connectors.ts` mounts `/api/connectors` with one endpoint: `GET /:userId/status` → `{ connectors, anyNeedsReauth }`. Session + ownership gated.

### Frontend
- `index.html` adds `#connectors-banner` sticky under the kill-switch banner.
- `styles.css`: amber (warning color, not red — re-auth is housekeeping, not panic); stacks below the autonomy banner when both fire.
- `app.js` `updateConnectorsBanner()` polls `/connectors/:userId/status` on every `navigate()` + every 60s. Conditional copy on `errorCode`. Single "Reconnect" CTA → `#/connect-gmail`.

## Verified

- `pnpm build` ✓
- `pnpm test` ✓ (713 API + 301 db + full monorepo green)

## Test plan

- [ ] Revoke SkyTwin's access at Google → next worker poll (~60s) writes `needs_reauth` to `connector_health`.
- [ ] `curl /api/connectors/<uuid>/status` returns `anyNeedsReauth: true`.
[...]

Issue #413, opened 2026-05-25
P1.13 FK constraints on users(id) lack ON DELETE CASCADE in 32 places (blocker of #376)


jayzalowitz:
**Parent:** #361
**Blocks:** #376

## Context

While verifying the cascade scope for #376 (P1.6 delete-my-data endpoint), the migration audit surfaced a structural gap: **the FK references on `users(id)` are inconsistent across the schema.** Migrations 015 onward (plus the migrations that came in 027+) correctly use `ON DELETE CASCADE`. Migrations 002-013 and the core `schema.sql` body do NOT — they use bare `REFERENCES users(id)`. This means a naive `DELETE FROM users WHERE id = ?` against the current schema fails with a foreign-key violation on the first child row.

This is a blocker for #376. Either #376 has to enumerate every table and DELETE in order (fragile — every new table added without CASCADE silently breaks #376), or this issue lands first and the cascade walk becomes a one-line operation.

## Root Cause Analysis

The schema grew organically. Early migrations (002-013) and the original `schema.sql` predate the project's adoption of `ON DELETE CASCADE` as the default for per-user tables. Each migration was authored independently; nobody enforced "if your table has a `user_id`, it needs `ON DELETE CASCADE`." Migration 015 (`ai-provider-settings.sql`) was the first to explicitly cascade, and migrations from then on followed the pattern — but the old refs were never backfilled.

CockroachDB does not retroactively cascade. You have to `ALTER TABLE ... DROP CONSTRAINT` and `ADD CONSTRAINT ... ON DELETE CASCADE`. Cheap, but it's 32 separate constraints across 13 files.

## Verified Current State

Verified 2026-05-25 against `jayzalowitz/launch-skytwin-demo @ 8bb5379`.

### Refs WITHOUT `ON DELETE CASCADE` (32 total)

| File | Lines | Table(s) |
|---|---|---|
| `packages/db/src/schemas/schema.sql` | 21, 36, 82, 101, 153, 169, 264 | 7 tables in the core schema (twin_profiles, etc. — full list to enumerate during fix) |
| `packages/db/src/migrations/002-oauth-tokens.sql` | 5, 18 | `oauth_tokens`, `oauth_states` |
| `packages/db/src/migrations/003-behavioral-patterns.sql` | 5, 21 | `behavioral_patterns`, `pattern_observations` |
| `packages/db/src/migrations/004-eval-history.sql` | 6, 21 | `eval_runs`, `eval_decisions` |
[...]

Issue #377, opened 2026-05-25
P1.7 OAuth re-auth has no user-facing surface


jayzalowitz:
**Parent:** #361

## Context

When a user revokes Gmail access at Google's side (Account → Security → Third-party apps → Remove SkyTwin) — or when their refresh token is otherwise invalidated by Google's policies (extended inactivity, password reset, security event) — the worker correctly classifies the failure as permanent and trips the per-user circuit breaker so it stops hammering Google's token endpoint. That part works.

What does not work: there is **no user-facing signal** that this happened. The dashboard renders as if everything is fine. The user sees their old decisions, their old learnings, an SSE "Listening" pill. They notice nothing wrong until a friend asks "did you get my email?" and they realize the twin hasn't processed anything in days.

This is the worst class of bug for a delegated-action product: **silent breakage that masquerades as normal operation**. The product appears to be working ("API is up, dashboard loads, no errors in DevTools") while every connector-dependent flow is dead. The user has no actionable next step because they don't know there's a problem.

## Root Cause Analysis

Three correctly-implemented mechanics that nevertheless add up to a silent-failure trap:

1. **Worker handles the permanent failure correctly, but only logs it.** `apps/worker/src/index.ts:236-246` catches `OAuthRefreshError` with `error.permanent === true`, force-trips the circuit breaker by recording three rapid failures (line 243), and returns early so the polling loop doesn't keep retrying a revoked token. There is no DB write, no event emission, no signal to the API. The only artifact of "Gmail is broken" is a log line nobody reads.

2. **`OAuthRefreshError.permanent` is correctly computed from the HTTP status.** `packages/connectors/src/oauth/google-oauth.ts:137-148` constructs the error: `permanent = statusCode === 400 || statusCode === 401 || statusCode === 403`. The Google `invalid_grant` response (which is what a revoked refresh token returns) is HTTP 400, so it is correctly classified as permanent. This part is right and should not change.

3. **No `/api/connectors/:userId/status` endpoint exists.** Confirmed by enumerating `apps/api/src/routes/` — 38 route files, none named `connectors.ts`, no `connectors/status` path string anywhere in `apps/web/public/js/`. The dashboard has no way to ask "is Gmail working?" so it cannot render a banner saying "no, it isn't."

The fix is one new DB table, one worker write on the permanent-failure path, one new API endpoint, and one dashboard banner. No core OAuth logic changes — that is already correct.

## Verified Current State

[...]

Issue #376, opened 2026-05-25
P1.6 No "delete my data" endpoint exists


jayzalowitz:
**Parent:** #361

## Context

A user who connects their Gmail, lets the twin learn for a week, and then decides "this isn't for me" has no way to delete what the product has accumulated about them. Twin profile, decision history, memory pages, knowledge triples, episodic memories, preferences, OAuth tokens, spend records, encrypted credential vault meta, idle-scan results — all of it remains, indefinitely, with no user-facing way to remove it short of `psql` and a list of 30+ tables.

This is the "GDPR Article 17 / right to erasure" gap, and it is independently a recovery gap. A user who accidentally trains a bad twin (wrong preferences, garbage signals from a stale connector, a domain enabled by mistake) cannot start over without a developer's help. A user who can't delete their data won't trust the app with new data.

It is also the kind of issue every public download visitor checks for in a privacy policy before installing — and SkyTwin's `docs/privacy.html` cannot truthfully say "we can delete your data on request" until the endpoint exists.

## Root Cause Analysis

Three structural reasons the delete path doesn't exist today:

1. **The users router never grew a DELETE.** `apps/api/src/routes/users.ts` has create (POST /), list (GET /), get-by-id (GET /:userId), update-trust-tier (PUT /:userId/trust-tier), update-domains (PUT /:userId/domains), seed-preferences (POST /:userId/seed-preferences) — and no DELETE handler. The route file was last extended for `seed-preferences` (#346 era); a DELETE was never added because the launch path didn't force it.

[...]

PR #425, opened 2026-05-26
p1.13: cascade ON DELETE on every legacy FK to users(id) (#413)

jayzalowitz:
## Summary

Pre-fix, 32 of 39 user-owned tables had a FK to `users(id)` without `ON DELETE CASCADE`. That meant any "delete my account" code path either had to enumerate every table manually (fragile) or hit FK violations on the first row in `behavioral_patterns` / `signals` / mempalace / `sessions`.

New migration `061-cascade-cleanup.sql` does the one-time DDL backfill:

\`\`\`sql
ALTER TABLE <t> DROP CONSTRAINT IF EXISTS <t>_user_id_fkey;
ALTER TABLE <t> ADD  CONSTRAINT <t>_user_id_fkey
  FOREIGN KEY (user_id) REFERENCES users(id) ON DELETE CASCADE;
\`\`\`

…for each of the 32 affected tables.

### Why CockroachDB's default constraint name works
The convention is `<table>_<column>_fkey` for any FK declared inline in a column definition without an explicit name. Verified across the codebase as the universal pattern — no FK to `users(id)` is hand-named (\`grep CONSTRAINT.*FOREIGN KEY packages/db/src/migrations/*.sql\` returns zero matches).

### Idempotency
- `DROP CONSTRAINT IF EXISTS` no-ops on a re-run
- `ADD CONSTRAINT` trips SQLSTATE `42710` (duplicate_object) which the migration runner's `IDEMPOTENT_DDL_CODES` set absorbs

### Safety net
New E2E test `cascade-cleanup.e2e.test.ts` (gated on `E2E=true`):
1. Queries `information_schema.referential_constraints` after migrations run, fails if any FK to `users(id)` is still `NO ACTION`
2. Exercises cascade end-to-end on `behavioral_patterns` (one of the 32) to prove the FK semantics — not just the metadata flag — actually delete child rows

If a fork ever adds a hand-named FK that the migration's conventional DROP would miss, the assertion test fires.

### Migration ordering
Number 061 leaves 060 to PR #422 (connector_health from #377). Both migrations are independent; either ordering is fine.

### Unlocks
- P1.6 #376 (delete-my-data endpoint) — without cascade, would have been a 30-statement manual purge or runtime FK error on first delete

## Test plan
- [x] `pnpm --filter @skytwin/db build` — clean
- [x] `pnpm --filter @skytwin/db test` — 301 pass, 19 skipped (no regressions)
- [ ] CI runs full test matrix
- [ ] (Manual) `E2E=true pnpm --filter @skytwin/db exec vitest run src/__tests__/cascade-cleanup.e2e.test.ts` against a CRDB with migrations applied


jayzalowitz:
Addressed both Copilot doc findings in 676a651: enumerated the cascading-FK categories accurately (mcp_*, capability_provenance_nodes, etc.) and fixed the "referencee" typo. CI should re-run.

Issue #396, opened 2026-05-25
P2.16 Trust-tier copy lacks promotion criteria


jayzalowitz:
**Parent:** #357

### Context
Settings explains the tiers in plain English. Does not explain "you'll be eligible for the next tier after N approvals AND Y hours."

### Scope
- Add a "What does it take to move up?" expandable section under the tier selector.
- Pull values from `PROMOTION_THRESHOLDS` so docs and code stay in sync.

### Acceptance criteria
- [ ] Expandable section visible.
- [ ] Values match `PROMOTION_THRESHOLDS`.

### Code pointer
`apps/web/public/js/pages/settings.js`, `packages/policy-engine/src/trust-tier-engine.ts`.

### Testing Plan

| Layer | What | Count |
|---|---|---|
| Vitest (jsdom) | Expandable section renders with values from `PROMOTION_THRESHOLDS` constant | +2 |
| Vitest (policy-engine) | Threshold export shape is stable (guards against accidental refactor breaking the UI) | +1 |
| Manual | Toggle expansion, change tiers, confirm correct thresholds shown per tier | +1 |

### Rollback Plan

Revert the PR. Pure UI addition; no schema, no policy-engine behavior changes. The `PROMOTION_THRESHOLDS` export added (if new) is harmless to leave behind, or can be removed in the revert.

### Effort Estimate

| Component | Hours |
|---|---|
| Export `PROMOTION_THRESHOLDS` from `policy-engine` (if not already) | 0.5h |
| Expandable section UI + copy | 1.5h |
| Tests | 0.5h |
| Manual QA pass | 0.5h |

**Total: ~3h.**

jayzalowitz:
Shipped via PR #427. Closing — code is in main.

PR #427, opened 2026-05-26
p2.16: show promotion criteria under tier selector (#396)

jayzalowitz:
## Summary

Settings explained the tiers but never said what it took to move from one to the next. Users only learned the floor existed when their next-tier offer didn't appear and they wondered why.

This adds an expandable section under the tier selector — \"What does it take to move up to <next tier>?\" — that renders three bullet lines for the gating criteria:

- N consecutive approvals (one rejection resets the streak)
- ≥X% approval ratio overall
- ≥Y hours in current tier (calibration takes time, not clicks)

Values are pulled from a \`PROMOTION_TIER_INFO\` constant in \`apps/web/public/js/pages/settings.js\` that mirrors \`PROMOTION_THRESHOLDS\` from \`packages/shared-types/src/policy.ts\` exactly.

For \`moderate_autonomy\` and \`high_autonomy\` (no automatic promotion), the section shows different copy explaining that the next jump (Full autopilot) is explicit opt-in.

### Safety net

\`packages/policy-engine/src/__tests__/promotion-thresholds-shape.test.ts\` locks the source-of-truth values (10/0.8/24h, 20/0.85/72h, 50/0.9/168h). If \`PROMOTION_THRESHOLDS\` drifts, the test fires pointing at the now-stale Settings copy in the same PR.

## Test plan
- [x] \`pnpm --filter @skytwin/policy-engine test\` — 182 pass (incl. 4 new)
- [ ] CI matrix
- [ ] (Manual) Open Settings, expand \"What does it take to move up?\" on observer / suggest / low_autonomy / moderate_autonomy users


jayzalowitz:
Addressed both Copilot findings in acc9d54: split the moderate_autonomy / high_autonomy fallback copy (high_autonomy IS Full autopilot, so the 'next jump' phrasing was wrong there), and replaced the dead-code `if (!t) return` guard in the shape-lock test with a loud `throw new Error` that both narrows for TS and fails the test with a useful message instead of silently passing on a missing entry.

Issue #394, opened 2026-05-25
P2.14 Loading state on capability-detail is three separate "Loading…" placeholders


jayzalowitz:
**Parent:** #357

### Context
Capability detail page renders three independent "Loading capability…" placeholders, one per data fetch. Looks janky.

### Scope
- Unify into a single skeleton with three visible sections.
- Show all three skeletons immediately; replace each with content as it loads.

### Acceptance criteria
- [ ] One skeleton, three sections; no flicker as data arrives.

### Code pointer
`apps/web/public/js/pages/capability-detail.js`.

### Testing Plan

| Layer | What | Count |
|---|---|---|
| Vitest (jsdom) | Initial render shows all three skeleton sections present | +1 |
| Vitest (jsdom) | Each section replaces its skeleton when its fetch resolves; siblings remain skeleton | +1 |
| Manual | Throttle network to Slow 3G in DevTools, confirm no flicker | +1 |

### Rollback Plan

Revert the PR. Pure frontend change — no server state, no migration. Worst case is reverting to the previous three-placeholder rendering.

### Effort Estimate

| Component | Hours |
|---|---|
| Single skeleton component with three slots | 1h |
| Wire three fetches to update slots independently | 1h |
| Tests | 0.5h |

**Total: ~2.5h.**

jayzalowitz:
Shipped via PR #429. Closing — code is in main.

PR #429, opened 2026-05-26
p2.14: single skeleton + parallel fetches on capability-detail (#394)

jayzalowitz:
## Summary

The capability-detail page painted \`Loading capability…\` then ran three fetches sequentially before snapping the full multi-card layout in place. End-to-end latency was the sum of three round-trips and the user saw a 1-line placeholder then a sudden 8-card paint.

Two changes:

1. **Three-section skeleton with shimmer** — renders immediately so the page structure is on-screen the instant the user lands. Header / Skills / Spending-guardrails stubs use a new \`.skeleton-pulse\` class with \`prefers-reduced-motion\` opt-out.
2. **Parallel fetches via \`Promise.allSettled\`** — end-to-end latency is now \`max(server, skills, policy)\` instead of the sum. Server fetch failure still routes to the error card (load-bearing for the rest of the render); skills/policy stay best-effort.

Reusable: the \`.skeleton-pulse\` CSS class can back future skeleton work elsewhere on the dashboard.

## Test plan
- [x] \`pnpm --filter @skytwin/web build\` — clean
- [ ] CI matrix
- [ ] (Manual) Throttle network to Slow 3G in DevTools, walk through a capability detail page, confirm skeleton → real content swap without flicker


jayzalowitz:
Both Copilot findings addressed in 3fb25b7. Server fetch is now awaited first on its own so its rejection surfaces immediately without waiting for slow/hung skills or policy fetches; the other two run in the background and are awaited after. JSDoc corrected to describe the actual single-shot swap behaviour (progressive per-section reveal is a follow-up).

Issue #392, opened 2026-05-25
P2.12 No metrics export for self-hosted operators


jayzalowitz:
**Parent:** #357

### Context
Metrics live in CRDB tables only. Self-hosters can't alert on circuit-breaker trips, LLM failure rates, or decision-pipeline latency without writing custom queries.

### Scope
- Add a Prometheus-format `/metrics` endpoint on the API.
- Expose: pool stats, circuit breaker state per provider, decision rate, approval rate, signal ingress rate, worker poll latency.
- Document in `docs/operations.md`.

### Acceptance criteria
- [ ] `curl http://localhost:3100/metrics` returns Prometheus exposition format.
- [ ] Grafana dashboard JSON committed to `docs/grafana/` for quick setup.

### Code pointer
`packages/observability/`, `apps/api/src/index.ts`.

### Testing Plan

| Layer | What | Count |
|---|---|---|
| Vitest (observability) | Exposition formatter: counter, gauge, histogram, labels, escape rules | +4 |
| Vitest (API) | `/metrics` route returns 200 with correct `Content-Type` and known series | +1 |
| Manual | Promtool lint of sample output; load the Grafana JSON in a fresh Grafana | +1 |

### Rollback Plan

Revert the PR. `/metrics` is a read-only endpoint — no schema changes, no writes. Any external Prometheus scraper will simply start returning 404 until the rollback is reverted again.

### Effort Estimate

| Component | Hours |
|---|---|
| Exposition formatter in `@skytwin/observability` | 2h |
| Register API route + content negotiation | 0.5h |
| Wire up existing counters/gauges (pool, breaker, decision, signal, worker) | 2h |
| Grafana dashboard JSON | 1h |
| `docs/operations.md` section | 0.5h |
| Tests | 1.5h |

**Total: ~7.5h.**

jayzalowitz:
Shipped via PR #431. Closing — code is in main.

Issue #378, opened 2026-05-25
P1.8 Worker has no DB-pool exhaustion / startup-failure handling


jayzalowitz:
**Parent:** #361

## Context

The first time SkyTwin meets a real outage in front of a real user, it will look like silence. The API will return 200 on `/api/health/live`, the SSE pill will say "Listening", and every actual user-facing operation will hang because the connection pool is exhausted and `pg.Pool.acquire()` has no timeout. The worker will exit cleanly after 30 seconds on startup if CRDB is unreachable, but it can't distinguish "DB is down" from "no users to discover yet" — and the API has no equivalent fail-fast at all.

This is the same failure mode that has cost every "looks healthy, isn't healthy" startup their first big incident. The product appears to be working in every shallow check while every deep operation hangs forever.

The fix is small and unsexy: a 5-second `acquireTimeoutMillis` on the pool, a `SELECT 1` startup probe in the API mirroring the worker's, and pool stats exposed on `/api/health/ready` so an operator (or a future canary) can actually see "we're 18/20 connections saturated" before the queue depth becomes a customer call.

## Root Cause Analysis

Three coordinated gaps that compound into the silent-failure trap:

1. **Pool has no `acquireTimeoutMillis`.** `packages/db/src/connection.ts:114-126` sets `max: 20, idleTimeoutMillis: 30_000, connectionTimeoutMillis: 5000`. There is **no** `acquireTimeoutMillis`. The `connectionTimeoutMillis` is the timeout for *establishing a new connection to the DB server*, not the timeout for *waiting on the pool* when all connections are in use. A 21st concurrent acquire waits forever — no error, no log, just a hanging promise.

2. **API has no startup probe.** `apps/api/src/index.ts` (full 348-line file) has no `SELECT 1` warmup, no startup hang timer, no exit-on-DB-unreachable behavior. It calls `app.listen(port, ...)` at line 273 and immediately serves traffic. If CRDB is unreachable, the API still binds the port, still returns 200 on `/api/health/live`, and only fails when a request hits a route that touches the DB — by which point the user is already confused.

[...]

PR #431, opened 2026-05-26
p2.12: Prometheus /metrics endpoint on API (#392)

jayzalowitz:
## Summary

Self-hosters had no way to alert on pool exhaustion or process health without writing custom CRDB queries. The metrics existed (\`getPoolStats\`, \`process.memoryUsage\`) but weren't exposed for scraping.

Three pieces:

1. **\`formatPrometheus\` helper in \`@skytwin/observability\`** — renders Prometheus text-format exposition. Hand-written, no \`prom-client\` dependency (we only need text exposition; one less dep for security-conscious self-hosters). Spec-compliant: HELP/TYPE/value triples, alphabetical label sorting, escape rules for backslash/quote/newline, \`NaN\` / \`+Inf\` / \`-Inf\` per wire format.

2. **\`GET /metrics\` route on the API** — exposes pg pool stats (\`skytwin_db_pool_total\`, \`_idle\`, \`_waiting\` — last is the canary for the #378 saturation class) plus process uptime / heap / RSS. Read-only and unauthenticated by design (scrapers don't carry sessions); zero per-user data in the payload.

3. **Docs** — \`docs/operations.md\` explains the endpoint + alert recommendations; \`docs/grafana/skytwin-overview.json\` is a 6-panel starter dashboard.

### Deferred to follow-up

Circuit-breaker state, decision rate, and worker poll latency live in the worker process and need a cross-process scrape strategy (IPC bridge or dual scrape config). Shipping pool + process now gives operators something to alert on today; the rest is additive.

### Tests
- 11 new in \`packages/observability/src/__tests__/prometheus.test.ts\` — wire-format compliance (HELP/TYPE/value, label sort, escape rules, NaN/Inf, empty payload, unit suffix)
- 6 new in \`apps/api/src/__tests__/metrics-route.test.ts\` — API surface (Content-Type, series names, pool-saturation passthrough, null-pool defaults)

## Test plan
- [x] \`pnpm build --concurrency=1 --force\` — clean
- [x] \`pnpm --filter @skytwin/observability test\` — passes
- [x] \`pnpm --filter @skytwin/api exec vitest run src/__tests__/metrics-route.test.ts\` — passes
- [ ] CI matrix
- [ ] (Manual) \`curl http://localhost:3100/metrics | promtool check metrics\` — should emit no warnings


jayzalowitz:
Both Copilot findings addressed in 2a57d2f. Test handler now mirrors all 7 production series (was missing heap_total_bytes and rss_bytes); added a series-catalog test that asserts every documented series is present so future drops/renames surface in CI rather than silently. Docs typo fixed.

Issue #393, opened 2026-05-25
P2.13 Decision-maker risk-assessment access lacks audit log


jayzalowitz:
**Parent:** #357

### Context
A worker that decrypts a user's tokens or queries a user's memory leaves no breadcrumb. No forensic trail for insider threat or lateral movement.

### Scope
- Add an `access_log` table: `(timestamp, userId, actor, action, resource_type, resource_id, request_id)`.
- Every token decryption logs an entry.
- Every cross-user query logs an entry.
- Settings page shows the user "your access log" (their own).

### Acceptance criteria
- [ ] Every credential-vault decryption writes an `access_log` row.
- [ ] Every API request that accesses sensitive data writes an `access_log` row.
- [ ] Settings page exposes the user's own log.
- [ ] Vitest: token decrypt → log row present.

### Code pointer
`packages/credential-vault/`, new migration for `access_log`.

### Testing Plan

| Layer | What | Count |
|---|---|---|
| Vitest (credential-vault) | Decrypt path writes `access_log` row with correct actor + resource_id | +2 |
| Vitest (API middleware) | Sensitive-route access writes row; non-sensitive routes do not | +2 |
| Vitest (repository) | Read-back filtered by userId; cross-user reads rejected | +2 |
| Manual | Trigger a decryption, view the row in Settings → Access Log | +1 |

### Rollback Plan

Revert the PR, then drop the new table to reclaim space:

```sql
DROP TABLE IF EXISTS access_log;
```

The migration is additive (new table only — no alterations to existing tables), so dropping it is safe. Any log rows written between deploy and rollback are lost; this is the intended behavior of a rollback. If audit data must be preserved across a rollback, `pg_dump` the table first before dropping.

### Effort Estimate

| Component | Hours |
|---|---|
| Migration: `access_log` table + indexes on (userId, timestamp) | 1.5h |
| Repository: `AccessLogRepository.write()` + `.listForUser()` | 1.5h |
| Write path in `credential-vault` (decrypt hook) | 1.5h |
| Write path in API middleware (sensitive-route tagging) | 2h |
| Read endpoint `/api/access-log` (user-scoped) | 1h |
| Settings page section + UI | 2h |
| Tests | 2.5h |

**Total: ~12h.**

jayzalowitz:
Shipped via PR #432. Closing — code is in main.

PR #432, opened 2026-05-26
p2.13: access-log foundation + credential-vault instrumentation (#393)

jayzalowitz:
## Summary

Pre-fix, a worker that decrypted a user's OAuth tokens left no breadcrumb — no forensic trail for insider threat or lateral movement, no way for the user to see "what touched my data."

Three pieces in this PR:

1. **\`access_log\` table** (migration \`062-access-log.sql\`) — \`(id, user_id, actor, action, resource_type, resource_id, request_id, occurred_at)\`. Indexed on \`(user_id, occurred_at DESC)\` so the future "your access log" Settings page is cheap. Append-only; cascades on user purge via the convention from #413.

2. **\`accessLogRepository\` in \`@skytwin/db\`** — \`.record()\` + \`.findByUser({ limit })\` (limit clamped to [1, 1000]).

3. **\`AuditLogPort\` interface in \`@skytwin/connectors\`** wired into \`DbTokenStore\` via \`setAuditLog(port, actor)\`. Every successful credential-vault decrypt emits one \`decrypt_oauth_token\` row. Plaintext-fallback paths (vault locked, no token) do NOT emit — those aren't privilege actions. Worker wires the sink with \`actor: 'worker'\`.

### Critical contract

The audit log is **fire-and-forget**. A logging failure NEVER blocks or denies a legitimate decrypt. Async + sync errors get caught + warn-logged at the call site.

### Tests
- **4 repo tests** in \`packages/db/src/__tests__/access-log-repository.test.ts\` — insert column order, optional-field coalescing, sort order, limit clamp
- **6 instrumentation tests** in \`packages/connectors/src/__tests__/db-token-store-audit.test.ts\` — emit on successful decrypt, no-op on plaintext fallback / null token / no port attached, swallow async + sync sink errors

### Follow-ups (same issue)

- API-route instrumentation for sensitive reads
- Settings → "Your access log" page

Shipped the worker-decryption path first because that's the highest-value forensic surface today.

## Test plan
- [x] \`pnpm build --concurrency=1\` — clean
- [x] \`pnpm --filter @skytwin/db exec vitest run src/__tests__/access-log-repository.test.ts\` — 4 pass
- [x] \`pnpm --filter @skytwin/connectors exec vitest run src/__tests__/db-token-store-audit.test.ts\` — 6 pass
- [ ] CI matrix


jayzalowitz:
All 3 Copilot findings addressed in 5cabc71. (1) Corrected the `record()` docstring — callers MUST catch the returned promise or use `.catch()`, pointed at the `DbTokenStore` pattern as the reference (the repo stays thin so future fail-loud callers can opt out of swallowing). (2) Extracted `sanitiseLimit()` that handles NaN / fractional / ±Infinity / negative inputs before SQL interpolation — added a 4-edge-case test. (3) Added optional `requestId?: string | null` to `AuditLogPort.recordAccess` so the port surface is uniform with the repository — today's worker-timer decrypts don't carry one but the future API-route instrumentation can.

Issue #391, opened 2026-05-25
P2.11 No "recent activity" log the user can scan


jayzalowitz:
**Parent:** #357

### Context
Decisions page is a list, not a stream. There's no "what did my twin do in the last hour" view with filters.

### Scope
- Add an Activity tab with a timeline view of every signal observed + decision made + action executed + feedback received.
- Filters: time range, signal source, decision outcome, risk tier.
- Click any item to drill into the explanation record.

### Acceptance criteria
- [ ] Activity tab shows last 24h by default.
- [ ] Filters work and update URL params.
- [ ] Drill-down to ExplanationRecord works.

### Code pointer
New page in `apps/web/public/js/pages/`, new API route in `apps/api/src/routes/`.

### Testing Plan

| Layer | What | Count |
|---|---|---|
| Vitest (API) | `/api/activity` route — pagination, filters, time-range bounds | +3 |
| Vitest (jsdom) | Filter state syncs with URL params + drill-down link | +2 |
| Manual | Generate fixture signals/decisions, walk the timeline + filters | +1 |

### Rollback Plan

Revert the PR. Read-only endpoint and a new page — no migration, no destructive writes. Existing decisions/signals data is untouched.

### Effort Estimate

| Component | Hours |
|---|---|
| `/api/activity` route (union query across signals + decisions + feedback) | 2h |
| Activity page + timeline rendering | 2h |
| Filter UI + URL param sync | 1h |
| Drill-down to ExplanationRecord | 1h |
| Tests | 1.5h |

**Total: ~7.5h.**

jayzalowitz:
Shipped via PR #433 (backend) + #434 (UI). Closing — code is in main.

PR #433, opened 2026-05-26
p2.11: recent-activity timeline endpoint (#391 backend slice)

jayzalowitz:
## Summary

The dashboard is missing an Activity tab. Users have to piece together \"what just happened\" across the Decisions page, the Approvals page, and SSE notifications. This PR ships the load-bearing **endpoint** the future tab will read from. UI + filters + drill-down link are tracked as follow-ups under the same issue.

## What's in this PR

\`GET /api/activity/:userId?hours=24&limit=200\` returns:

\`\`\`json
{
  \"events\": [
    { \"id\": \"fb:fb-1\", \"kind\": \"feedback\", \"at\": \"...\", \"summary\": \"...\", \"decisionId\": \"...\" },
    { \"id\": \"dec:dec-1\", \"kind\": \"decision\", \"at\": \"...\", \"summary\": \"...\", \"domain\": \"...\", \"decisionId\": \"dec-1\" },
    { \"id\": \"sig:sig-1\", \"kind\": \"signal\", \"at\": \"...\", \"summary\": \"...\", \"domain\": \"...\" }
  ],
  \"total\": 3, \"hours\": 24, \"limit\": 200
}
\`\`\`

- Three underlying repository fetches run in **parallel** via \`Promise.allSettled\` — a CRDB blip on one source returns partial results (better to show 2 of 3 timelines than fail the whole tab).
- \`decisionId\` on feedback + decision events lets the future UI drill into the ExplanationRecord with no extra lookup.
- Bounds: \`hours\` clamped to \`[1, 720]\` (30d), \`limit\` clamped to \`[1, 500]\`; malformed query params fall back to defaults.

### Auth
Existing \`sessionAuth + requireOwnership + bindUserIdParamValidator\`. Malformed userId rejects with \`400 invalid_user_id\` before any repo runs.

### Tests

7 new in \`apps/api/src/__tests__/activity-route.test.ts\` — empty-user default, newest-first merge, lookback-window drop, hours/limit clamp on both ends, partial-result resilience, malformed-userId rejection, row-cap on oversized union.

## Test plan
- [x] \`pnpm --filter @skytwin/api build\` — clean
- [x] \`pnpm --filter @skytwin/api exec vitest run src/__tests__/activity-route.test.ts\` — 7 pass
- [ ] CI matrix
- [ ] (Manual) \`curl http://localhost:3100/api/activity/<uuid>?hours=24\` — observe shape


jayzalowitz:
All 3 Copilot findings addressed in ef9dbd7. (1) Added a defensive in-route slice that bounds each source's input to the merge buffer before union — keeps memory bounded for heavy users with wide windows; SQL-side LIMIT in signalRepository.getRecent is a follow-up. (2) Pre-compute timestamp ms once per row into an internal `_ts` field; filter + sort use the cached number; `_ts` is stripped from the wire payload before serialisation so the API shape is unchanged. (3) Tightened the `decisionId` doc — populated on decision/feedback events, NEVER on signal (signals don't carry a decision link today). All 7 route tests still pass.

Issue #397, opened 2026-05-25
P2.17 No demo script in `docs/`


jayzalowitz:
**Parent:** #357

### Context
`docs/launch-plan.md` Tier 1.4 calls for a video. There's no `docs/demo.md` for "if you have 5 minutes and a friend, do these 7 things."

### Scope
- Write `docs/demo.md` with a 5-minute happy path:
  1. Launch app (or click sample profile)
  2. Show empty dashboard + "what's coming"
  3. Connect Gmail → first signal arrives → first decision appears
  4. Approve a decision; explain the explanation
  5. Reject a decision; explain how feedback flows back
  6. Visit Settings; show trust tier + spend cap
  7. Hit the "pause" button; show how to undo

### Acceptance criteria
- [ ] `docs/demo.md` exists, 5-min readable.
- [ ] Each step has a screenshot.
- [ ] Demo video uses this script.

### Testing Plan

| Layer | What | Count |
|---|---|---|
| Manual review | Walk the 7 steps against a fresh dev install; time it; confirm each step works as written | +1 |
| Manual review | Screenshots match current UI (no stale chrome) | +1 |
| Link check | `pnpm` doc link checker (or manual) catches dead anchors and broken image refs | +1 |

### Rollback Plan

Revert the PR. Docs-only change; deletes `docs/demo.md` and any committed screenshots under `docs/img/demo/`. No code, no schema, no external state.

### Effort Estimate

| Component | Hours |
|---|---|
| Write the 7-step script + intro/outro | 2h |
| Capture screenshots (7 frames) | 1h |
| Polish + dry-run timing | 0.5h |
| Link check + final review | 0.5h |

**Total: ~4h.**

jayzalowitz:
Shipped via PR #449. Closing — code is in main.

PR #449, opened 2026-05-26
p2.17: demo script — five minutes, seven steps (#397)

jayzalowitz:
## Summary

Tier 1.4 of \`docs/launch-plan.md\` called for a launch video. There was no script. This adds \`docs/demo.md\` with the seven-step walkthrough the launch-video script needs:

1. Cold-load the dashboard (30s)
2. Orient on the empty state (45s)
3. Connect Gmail or use the sample profile (45s)
4. Approve a decision; explain the explanation (60s)
5. Reject a decision; show feedback flows back (45s)
6. Settings — trust tier + spend cap + delete-my-data (45s)
7. Pause everything; the trust moment (30s)

Each step has a goal, a target time, and an explicit \"what this proves\" callout the reader can use as the talking point. Includes a time-budget table (5:00 total) with compression guidance for steps to shorten if running long, plus a \"recording the launch video\" section with the practical bits (1280×800 Chrome window, sample profile not real Gmail, headset mic, hard cut at 5 min).

### Screenshots

Intentionally absent from this PR. The text walkthrough is independently runnable against a fresh dev install, and the screenshots are tracked as a follow-up so they get captured against the actual UI rather than an early-mockup that's already drifted. This file doubles as the launch-video script.

## Test plan
- [ ] CI matrix (markdown lint)
- [ ] (Manual) Time the walkthrough end-to-end against a fresh dev install with the sample profile; confirm it lands close to 5:00


jayzalowitz:
All 5 Copilot findings addressed in 3f0c256 — every one was real factual drift between the script and the code/repo. Verified each against source before fixing:

1+5. `docs/img/demo/` doesn't exist; the repo's demo stills live under `docs/screenshots/`. Pointed at the right directory in both the doc and the CHANGELOG. Reframed the follow-up as "fresh captures against the post-Tier-2 UI" since the baseline already exists.
2. "5 sample decisions" was wrong (`grep` on the seed shows 8 today). Reworded to "a handful across several domains" + an explicit note that the count moves as we tune the demo data.
3. "Listening — Gmail connected" / "Idle" strings don't exist. Real renderer strings (from `apps/web/public/js/app.js`) are "Listening" / "Connected" / "Reconnecting…" / grey idle dot. Rewrote the paragraph to match + cited the file.
4. Spend cap defaults were stated as $10/$50; actual defaults from `settings.js` are 10000 cents = $100/action and 50000 cents = $500/day. Fixed + cited the file.

Pure doc-accuracy fixes. Script now matches what a reader actually sees on screen.

Issue #389, opened 2026-05-25
P2.9 "Computer" onboarding path is a no-op stub


jayzalowitz:
**Parent:** #357

### Context
`apps/api/src/routes/onboarding.ts:341,361` accepts the "Let SkyTwin learn from your computer" choice and shows a recipe preview, but never actually enables the idle miner. User completes onboarding with zero signals enabled.

### Scope
Either:
- (a) Ship the enable endpoint that actually wires up the idle miner.
- (b) Hide the button with a "Coming soon" badge until (a) ships.

Recommendation: **(a)** — idle miner code already exists in `@skytwin/idle-miner` (see `packages/idle-miner/src/miner.ts`, `idle-detector.ts`, `runtime.ts`), just needs wiring.

**Note on integration with #382:** the desktop idle bridge at `apps/desktop/src/idle-bridge.ts` is the runtime trigger for when the miner should actually run (scan during idle, pause when user returns). #382 makes the idle bridge no longer dead code; this issue depends on that wiring to be useful. Land #382 first or land both together — if (a) ships before #382, the miner will start on a timer instead of on idle, which is acceptable but wasteful.

### Acceptance criteria
- [ ] Clicking "Let SkyTwin learn from your computer" enables the idle miner and shows a confirmation.
- [ ] Settings has a toggle to disable later.
- [ ] First signals appear in the dashboard within 5 minutes of idle detection.

### Code pointer
- Wire `apps/api/src/routes/onboarding.ts:341,361` to call into `packages/idle-miner/` (specifically `miner.ts` / `runtime.ts`) instead of stubbing.
- Runtime trigger lives in `apps/desktop/src/idle-bridge.ts` (currently unused — see #382).
- Settings toggle: add to existing settings page alongside other connector toggles.

### Testing Plan

| Layer | What | Count |
|---|---|---|
| Unit (vitest) | Onboarding route: POST enable-idle-miner → persists setting + starts miner runtime. POST disable → stops miner. Idempotent (enable twice = no error). | +3 |
| Integration | Onboarding click → route hit → idle miner runtime started → first signal emitted within mocked idle window. Verify signal carries correct `AuthoringTier` and provenance. | +2 |
| Manual smoke | Fresh install, complete onboarding with "Let SkyTwin learn from your computer", wait/simulate idle, verify signal appears in dashboard within 5min. | +1 manual |

### Rollback Plan

[...]

Issue #382, opened 2026-05-25
P2.2 Idle bridge detects idle but nothing uses the signal


jayzalowitz:
**Parent:** #357

### Context
`apps/desktop/src/idle-bridge.ts` detects idle state and dispatches `idle-state-changed` events at `main.ts:148`. The web frontend never subscribes. The API never receives the state. Dead code.

### Scope
Either:
- (a) **Use it:** pause connector polling and decision generation during idle, resume on active.
- (b) **Remove it:** delete the dead code.

Recommendation: (a) — pausing background work when the user walks away is a real battery win.

### Acceptance criteria
- [ ] Idle for >5min → worker pauses polling.
- [ ] Active → worker resumes within 1 cycle.
- [ ] Settings toggle: "Pause background work when idle" (default ON).

### Code pointer
`apps/desktop/src/idle-bridge.ts`, `apps/desktop/src/main.ts:148`.

### Testing Plan

| Layer | What | Count |
|---|---|---|
| Unit (vitest) | Worker poller `pause()` / `resume()` state transitions; settings toggle gates the pause path. | +3 |
| Integration | Simulated `idle-state-changed` event → worker enters paused state → simulated `active` → resumes within one cycle. Mock the bridge IPC. | +1 |
| Manual smoke | Real idle (5min AFK) on macOS, observe worker logs go quiet, return → polling resumes. | +1 manual |

### Rollback Plan

Revert the PR. The settings toggle persists in user settings JSON; on revert the field becomes unused and is ignored (no migration needed). If shipped to users, leave the toggle field in the schema as a deprecated no-op for one release before cleanup.

### Effort Estimate

| Component | Hours |
|---|---|
| IPC bridge: desktop idle event → API/worker | 1.5h |
| Worker pause/resume state machine + poller integration | 2h |
| Settings UI toggle + persistence | 1h |
| Tests + manual smoke | 1.5h |

**Total: ~6h.**

PR #450, opened 2026-05-26
p2.9: "learn from your computer" → Coming soon (#389 option b)

jayzalowitz:
## Summary

Pre-fix, clicking the third welcome-screen choice in the first-run wizard transitioned through \`computer_choice\` → \`recipe_preview\` and dropped the user on the dashboard with **zero signals enabled** — the \`idle_miner_poll\` step doesn't actually enable the miner, and the desktop idle bridge in \`apps/desktop/src/idle-bridge.ts\` is dead code today (tracked by #382). Users who picked this path were silently routed to a no-op and then wondered why nothing was happening.

Per #389 acceptance-criteria option (b), the button is now:

- \`disabled\` + \`aria-disabled=\"true\"\` so it can't be clicked
- styled with \`opacity 0.55\` + \`cursor: not-allowed\`
- tagged with a \"Coming soon\" badge inline with the title
- hover tooltip explaining the state
- copy links to the tracking issue (#389) so wiring status is visible

Email + about-me paths are unchanged.

Option (a) — actually wiring up the idle miner — is the same issue's primary fix and depends on #382 (idle-bridge wire-up) landing first. This is the safe interim that keeps cold-loaders from completing onboarding into a silent dead-end.

## Test plan
- [x] \`pnpm --filter @skytwin/web build\` — clean
- [ ] CI matrix
- [ ] (Manual) Open the first-run wizard cold, confirm the computer button is greyed + non-clickable, hover tooltip shows, badge reads \"Coming soon\"


jayzalowitz:
Addressed in f5c37b8. Moved the #389 link out of the disabled button into a sibling `<p>` tied via `aria-describedby` — screen readers still surface the "see #389 for status" hint when the button gets focus, and the link is now actually clickable (a disabled `<button>` blocks pointer events on its children, so the link wouldn't have worked anyway). Button stays disabled + greyed.

Issue #414, opened 2026-05-25
P0.10 Build deterministic demo-video recorder (unblocks #351 Google submission)


jayzalowitz:
**Parent:** #359
**Blocks:** #351 (Google submission needs the demo video)

## Context

Google's restricted-scope OAuth verification (#351) requires a 2-3 minute demo video showing install → consent → first signal → first decision → undo. Without the video, the submission is incomplete; with the video, Google reviews on a 4-6 week SLA. The video is on the critical path for the launch criterion "real strangers can install without configuring their own OAuth client."

Today this work is in the "you record video" column — a human-only task that depends on a human walking through the app on a clean install, with a microphone, and editing the result. That introduces a 4-8 hour human bottleneck and a quality-variance risk (cuts won't be perfectly consistent across re-takes).

We already have a working solution in a sibling codebase: `~/depobot/.conductor/las-vegas-v1/scripts/demo/deposition_prep/` produces frame-perfect demo videos using Selenium screenshots @ 1 fps + ElevenLabs narration + ffmpeg mux. Adapting that pipeline to SkyTwin removes the human bottleneck — the script regenerates the video deterministically every time the app changes.

## Why now

- #351 cannot ship without the video. The video is the long-pole external dep on launch.
- Re-recording any time the UI changes (which it will, repeatedly, through Epic A and Epic D) is cheap if scripted, painful if manual.
- The demo video is also reusable for launch announcement (HN post, README, social), so it earns its build cost beyond Google submission.

## Verified Current State

Verified 2026-05-25.

| What | Where | Status |
|---|---|---|
| ElevenLabs API key | User has one (used in depobot) | Available; placement TBD per implementation |
| Demo recorder source template | `~/depobot/.conductor/las-vegas-v1/scripts/demo/deposition_prep/` | Working — produces `~/Desktop/depo_prep_demo.mp4` |
| ffmpeg | homebrew, depobot uses it | Installed |
| ChromeDriver + headless Chrome | depobot uses it | Installed |
| BlackHole 2ch (system audio capture) | depobot prerequisite | Likely installed; only needed if capturing app audio (vs narrator-only) |
| Selenium + Pillow + webdriver-manager | depobot's `backend/venv` | Reusable or duplicate into a skytwin venv |
| SkyTwin demo flow | Depends on Epic A (#358) shipping first | **Blocked until #362-#367 land** |

### What the depobot pipeline already does well (reusable)

[...]

Issue #351, opened 2026-05-22
Submit bundled OAuth client for Gmail restricted-scope verification


jayzalowitz:
**Parent:** #359

## Context

SkyTwin uses a tiered OAuth design today (full spec in [docs/google-verification.md](https://github.com/jayzalowitz/skytwin/blob/main/docs/google-verification.md)):

- **Tier 1 — bundled client (Identity + Calendar):** The SkyTwin-team OAuth client `594829999930-kpjopcs1pak0rp0omimuegr5ugcv5l8h.apps.googleusercontent.com` ships in `apps/desktop/src/service-manager.ts`. Verified by Google for sensitive but not restricted scopes; no third-party assessment fee. Used by default for every "Sign in with Google" click.
- **Tier 2 — BYO Gmail:** Users who want Gmail features follow the 5-minute walkthrough at https://jayzalowitz.github.io/skytwin/connect-gmail.html to create their own Google Cloud OAuth client and paste it into SkyTwin's Setup page. Their own client is exempt from app verification (it's a developer using their own credentials), so Gmail's restricted scopes work immediately.

This split is exactly right for launch — $0 cost, every user is unblocked. But the 5-minute BYO step is real friction. This issue tracks the eventual work to remove it.

## What needs to happen

- [ ] **Pick a CASA assessor.** Tier 2 (read-only) for `gmail.readonly` alone, or Tier 3 (modify) for `gmail.readonly` + `gmail.modify`. List of empanelled assessors: https://cloud.google.com/security/compliance/casa-tier-2-assessors. Bishop Fox, Leviathan, Schellman, NCC are common picks.
- [ ] **Pre-assessment readiness.** Walk the CASA checklist against the data-flow in [docs/technical-spec.md](https://github.com/jayzalowitz/skytwin/blob/main/docs/technical-spec.md), the encryption in [packages/credential-vault](https://github.com/jayzalowitz/skytwin/tree/main/packages/credential-vault), and the API in [apps/api](https://github.com/jayzalowitz/skytwin/tree/main/apps/api). Fix anything obvious before paying for the assessor's time.
- [ ] **Engage the assessor.** Budget: 4–8 weeks engagement; \$15k–\$30k (Tier 2) or \$25k–\$50k (Tier 3) per cycle. The fee renews **annually** regardless of whether SkyTwin changed.
[...]

PR #454, opened 2026-05-27
p0.10: deterministic demo-video recorder — Playwright + Piper TTS + ffmpeg (#414)

jayzalowitz:
## Summary

Builds the 5-minute launch video that the Google OAuth restricted-scope review (#351) is waiting on. **Every dependency is local-first and open-source** — Playwright instead of Selenium, Piper TTS via the existing \`/api/voice/synthesize\` endpoint instead of ElevenLabs, \`ffmpeg\` for assembly. Zero new API keys, zero new runtime services. The Piper voice model is the same binary already shipped via \`@skytwin/embedded-llm\` for the on-device voice loop.

## What's in this PR

New \`scripts/demo/\` standalone package:

| File | Role |
|---|---|
| \`timeline.json\` | Cue points + Playwright actions per step, mirroring the 7-step walkthrough in \`docs/demo.md\` |
| \`src/timeline.ts\` | Strict schema validator — a typo fails at load time, never produces a broken mp4 |
| \`src/record.ts\` | Playwright Chromium driver, pre-seeds sample-profile userId, parallel screenshot loop at 1 fps |
| \`src/narrator.ts\` | Piper TTS caller, sha256-cached by \`(voice, text)\` so re-runs after copy edits are ~5s |
| \`src/assemble.ts\` | ffmpeg orchestration: silent H.264 from frames + \`adelay\`+\`amix\` to position narration at each step's cumulative cue offset |
| \`README.md\` | Setup, recording, editing the script, sub-commands, environment knobs, why-not-Selenium-or-ElevenLabs |

### Top-level pnpm shortcuts

\`\`\`sh
pnpm demo:install     # one-time setup (Playwright + Chromium binary)
pnpm demo:record      # full pipeline
pnpm demo:narrate     # TTS only (smoke-test Piper, find copy bugs early)
\`\`\`

### Pre-flight

The recorder checks the dev server is reachable + the sample profile is seeded **before** launching Chromium, so a forgotten \`pnpm dev\` fails in seconds rather than after a 60-second screenshot loop.

### Determinism

Output is byte-deterministic given:

1. Same \`timeline.json\`
2. Same Piper voice model file (the embedded-llm dep)
3. Same seeded sample profile (the dashboard's content is what varies most across runs — the recorder pre-seeds the sample userId via Playwright's \`addInitScript\` rather than letting a real Gmail signal stream drift the on-screen state)
4. Same Playwright + Chromium minor versions (pinned via the standalone package's lockfile)

### Why not in the workspace?

\`scripts/demo/\` is deliberately NOT a workspace package (uses \`--ignore-workspace\` on install) so its Playwright dev-dep doesn't pollute the monorepo's install graph. It's a bench tool, not part of the build graph.

## Test plan
- [x] \`cd scripts/demo && pnpm install --ignore-workspace && pnpm exec tsc --noEmit\` — clean typecheck against the real Playwright + Node types
- [x] Root \`pnpm install\` still works (the new top-level scripts don't disturb the workspace)
- [ ] CI matrix
- [ ] (Manual) \`pnpm dev\` + \`pnpm db:seed\` in two terminals, then \`pnpm demo:record\` — confirm \`scripts/demo/out/demo.mp4\` renders the 5-minute walkthrough end-to-end
- [ ] (Manual) Edit a \`narration\` line in \`timeline.json\`, re-run, confirm only the edited line re-synthesises and the WAV cache reuses the rest


jayzalowitz:
Addressed all 9 Copilot findings in 5da8ac1:

**Real bugs (2)**
1. `captureLoop` now threads a global `startIndex` — previously each step rewrote `frame-000000.png`, clobbering prior segments.
2. `captureLoop` keeps capturing until BOTH `durationSeconds` has elapsed AND the actions promise has settled, so a slow Playwright selector no longer leaves a hole at the end of a segment (and the post-loop `await actionPromise` still re-raises any real failure).

**Polish (7)**
3. `record.ts` pre-seed comment now describes what we actually set (only `skytwin_userId`) and points at the onboarding source of truth.
4. `parseTimeline()` rejects duplicate step ids — they'd silently clobber each other in `stepStartTimes`.
5. Removed the dangling `$schema` reference from `timeline.json`; the runtime validator in `src/timeline.ts` is the schema.
6. `narrator.ts` comment no longer references the non-existent `--user-id` flag; documents the env var instead.
7. Narration cache filename is now `<hash>.wav` (hash-only) — renaming a step no longer forces re-synthesis.
8. + 9. `assemble.ts` got a real CLI entrypoint; `pnpm assemble` now actually re-muxes `out/frames` + `out/narration-cache` → `out/demo.mp4` instead of being a silent no-op. README updated.

Typecheck clean.

Issue #381, opened 2026-05-25
P2.1 Desktop "close window" doesn't quit, no tutorial for tray


jayzalowitz:
**Parent:** #357

### Context
Closing the desktop window hides it (not quits). App keeps running in tray. Users hit close, assume the app is gone, but it's silently consuming battery.

### Scope
- One-time toast on first window close: "SkyTwin keeps running so it can act on signals. Quit fully from the menu bar icon."
- Add a "Quit SkyTwin" item to the tray menu (verify it's there).
- Document tray icon in README screenshots.

### Acceptance criteria
- [ ] First close → toast appears.
- [ ] Toast dismissible, doesn't reappear in same session.
- [ ] Tray menu has clear "Quit SkyTwin" item.

### Code pointer
`apps/desktop/src/main.ts` close handler.

### Testing Plan

| Layer | What | Count |
|---|---|---|
| Unit (vitest) | "First close in session shows toast; subsequent closes don't" — extract `shouldShowFirstCloseToast(sessionState)` helper and test the state machine. | +2 |
| Manual smoke | Build the desktop app, close window on macOS + Windows, verify toast + tray "Quit SkyTwin" item works. Electron tray behavior is platform-specific and not worth a Spectron harness for one menu item. | +1 manual |

### Rollback Plan

Revert the PR. No data or schema implications — toast suppression is in-memory session state only.

### Effort Estimate

| Component | Hours |
|---|---|
| Toast component + first-close session state | 1h |
| Tray menu "Quit SkyTwin" item + verify on macOS/Windows/Linux | 1h |
| README screenshot of tray icon | 0.5h |
| Manual smoke across platforms | 0.5h |

**Total: ~3h.**

jayzalowitz:
Shipped via PR #455 (squash-merged as d4ffdf4). First window-close in a session now fires a tray-explainer toast via the existing showToast helper, with platform-aware wording ('menu bar icon' on darwin, 'system tray icon' elsewhere). Manual smoke on packaged builds across macOS/Windows/Linux is deferred to a packaged-build pass.

PR #455, opened 2026-05-27
p2.1: first window-close toast explaining the tray (#381)

jayzalowitz:
## Summary

Closes the discoverability gap behind #381 P2.1: closing the desktop window hides to tray (correct background-daemon behaviour) but users who reached for ⌘W saw the window vanish and assumed the app was dead. The Quit affordance is already in the tray menu; we just weren't telling the user.

- **`apps/desktop/src/first-close-toast.ts` (new)** — pure state-machine helper. `shouldShowFirstCloseToast(state)` returns true once per session, flips the state.
- **`apps/desktop/src/main.ts`** — close handler calls the helper, sends `show-first-close-toast` IPC on the first hit, hides the window.
- **`apps/desktop/src/preload.ts`** — new `onFirstCloseToast(listener)` API.
- **`apps/web/public/js/app.js`** — subscribes and shows an 8s info toast via the existing `showToast()` helper.
- **3 new vitest cases** lock the state machine.

Suppression is session-scoped per the issue's AC. Cross-session re-shows are intentional — the hint should resurface for someone who quits and relaunches having forgotten the tray semantics.

## Test plan

- [x] `pnpm --filter skytwin-desktop test` — 203 pass (3 new for first-close-toast)
- [x] `pnpm --filter skytwin-desktop exec tsc --noEmit` — clean
- [ ] Manual smoke on packaged macOS / Windows / Linux builds — deferred to a packaged-build pass (Conductor harness can't drive a real Electron window)


jayzalowitz:
Addressed all 3 Copilot findings in 348578f:

1. `main.ts` — destroyed-webContents guard now runs **before** `shouldShowFirstCloseToast()` so the one-shot flag doesn't flip on a dispatch that never happens. Added try/catch around `send()` in case the renderer crashes between the check and the send.
2. `app.js` — toast wording branches on `window.skytwinDesktop.platform`: "menu bar icon" on darwin, "system tray icon" on Windows/Linux.
3. `first-close-toast.ts` — module docstring no longer frames close-to-tray as macOS-specific; the close handler is unconditional. Calls out that platform vocabulary lives in the renderer.

Typecheck + tests green.

PR #457, opened 2026-05-29
p2.2: idle bridge pauses the worker when the user walks away (#382)

### LINKED ISSUE #382: P2.2 Idle bridge detects idle but nothing uses the signal

jayzalowitz:
## Summary

Closes #382 P2.2. The idle bridge already detected idle/active (via `powerMonitor`) but fired into a console log + an unconsumed IPC — dead code. This PR connects it to `ServiceManager.pause()/resume()` so the worker stops polling Gmail / generating decisions when the user is away, and resumes when they return.

- **`apps/desktop/src/idle-pause-controller.ts` (new)** — side-effect-injected state machine. Pauses on idle, resumes on active **only if it was the one that auto-paused** (so a manual kill-switch pause survives an idle cycle). Short-circuits when the setting is off; disabling mid-auto-pause resumes immediately; `onManualPauseChange()` clears the auto flag so the user's manual intent always wins.
- **`apps/desktop/src/desktop-preferences.ts` (new)** — electron-store-backed `idlePauseEnabled` (default ON).
- **`main.ts`** — wires the controller to `serviceManager`; idle handler forwards transitions; `pause-twin`/`resume-twin` IPC call `onManualPauseChange()`; new `get/set-idle-pause-enabled` IPC.
- **`preload.ts`** — `getIdlePauseEnabled` / `setIdlePauseEnabled`.
- **`settings.js`** — Desktop-section toggle "Pause background work when idle", round-tripped + toast on save.
- **11 new vitest cases** cover the controller transitions.

## Acceptance criteria

- [x] Idle >5min → worker pauses (bridge threshold is 300s; controller calls `pause()`)
- [x] Active → worker resumes (within one idle-bridge poll cycle)
- [x] Settings toggle "Pause background work when idle" (default ON)

## Test plan

- [x] `pnpm --filter skytwin-desktop test` — 222 pass (11 new)
- [x] `pnpm --filter skytwin-desktop exec tsc --noEmit` — clean
- [x] `node --check apps/web/public/js/pages/settings.js`
- [ ] Manual 5-min-AFK smoke on a packaged build — deferred to a packaged-build pass (harness can't drive `powerMonitor`)


jayzalowitz:
Addressed the concurrency finding in 1421631.

`IdlePauseController` now serializes all pause/resume work through an internal promise queue. Both `onIdleStateChange` and `onEnabledChanged` chain onto it, so the state decision **and** the async side-effect run atomically — a slow `pause()` (waiting for the worker to exit) can no longer be interleaved with a `resume()` that would orphan the freshly-started worker. The chain tail recovers from a thrown side-effect so one failure can't wedge subsequent transitions. `onManualPauseChange` stays synchronous (it only clears a flag).

2 new tests: (1) a slow pause followed immediately by a resume runs strictly in order; (2) a thrown `pauseServices` doesn't wedge the next transition. 224 desktop tests pass; tsc clean.

Issue #387, opened 2026-05-25
P2.7 Mobile notifications open the app, not the specific approval


jayzalowitz:
**Parent:** #360

### Context
Approval notification → tap → app opens to the Approvals tab root (not the specific approval). User has to scroll to find it.

### Scope
- Add deep-link routing: `skytwin://approvals/{id}` opens the specific approval.
- Add custom notification actions: "Approve" / "Reject" inline from the notification (iOS Notification Service Extension + Android NotificationCompat actions).

### Acceptance criteria
- [ ] Tap notification → app opens to the specific approval, scrolled into view.
- [ ] Long-press notification → "Approve" / "Reject" actions visible.
- [ ] Approving via notification action submits and dismisses without opening the app.

### Code pointer
`apps/mobile/src/services/notifications.ts`.

### Testing Plan

| Layer | What | Count |
|---|---|---|
| Unit (vitest) | Deep-link parser: `parseSkytwinUrl('skytwin://approvals/abc-123')` → `{ route: 'approval-detail', id: 'abc-123' }`. Cover invalid URLs, missing id, unknown route. | +4 |
| Manual smoke | iOS: send push, tap notification, verify deep-link lands on approval detail. Long-press, tap "Approve", verify submission without app focus. Same on Android. Native notification action handlers are not unit-testable cheaply. | +2 manual |

### Rollback Plan

Revert the PR. No server-side data implications. Notification action handlers register at app startup; on revert, old notifications without action buttons reappear. iOS Notification Service Extension is a separate target — confirm `Info.plist` and entitlements changes are also reverted.

### Effort Estimate

| Component | Hours |
|---|---|
| Deep-link URL scheme registration + route handler | 1.5h |
| Approval-detail screen scroll-to-approval-id logic | 1h |
| iOS Notification Service Extension + action handlers | 2.5h |
| Android NotificationCompat actions + background submit | 2h |
| Manual cross-platform smoke | 1h |

**Total: ~8h.**

jayzalowitz:
Deep-link routing slice shipped via PR #459 (squash-merged): tapping an approval push now opens the specific approval (expanded + scrolled into view) via the `skytwin://approvals/<id>` scheme, with warm + cold-start routing and the SSE handler stamping the approval id onto each notification. 

[...]

Issue #386, opened 2026-05-25
P2.6 Mobile voice upload is base64 full-buffer, no retry


jayzalowitz:
**Parent:** #360

### Context
Voice recording is base64-encoded and uploaded as a single POST to the desktop API. A 2-minute recording is ~3MB wire. On cellular with spotty coverage, the connection drops mid-upload with no retry and no visible progress.

### Scope
- Chunk the upload into 256KB pieces.
- Server reassembles via a session token.
- On disconnect: retry the failed chunk, not the whole upload.
- UI: progress bar during upload, clear "Connection lost — retrying" state.

### Acceptance criteria
- [ ] 2-minute voice memo uploads with visible progress.
- [ ] Disconnect mid-upload → resume from the next chunk after reconnect.
- [ ] Cancel button works mid-upload.

### Code pointer
`apps/mobile/src/services/voice-service.ts`, `apps/api/src/routes/voice.ts`.

### Testing Plan

| Layer | What | Count |
|---|---|---|
| Unit (vitest) | Chunker: splits buffer into 256KB pieces with correct offsets, last chunk handles remainder. Reassembler: out-of-order arrival reassembles correctly, missing chunk returns gap list. | +4 |
| Integration | Server upload session: open → upload chunks 1,2,3 → finalize. Failure case: open → upload 1,2 → reconnect → upload 3 → finalize. Stale-session cleanup after timeout. | +3 |
| Manual smoke | Record 2-min voice memo on cellular, toggle airplane mode mid-upload, verify resume + transcript matches. | +1 manual |

### Rollback Plan

Revert the PR. **Has server-side state implications:** chunked-upload sessions live in a session store (in-memory Map keyed by session token, with TTL ~10min). On revert, in-flight sessions drop; clients see upload failures and re-record. No persistent data — sessions are ephemeral by design.

### Effort Estimate

| Component | Hours |
|---|---|
| Mobile chunker + retry logic + session token handling | 2.5h |
| API: session-open / chunk-upload / finalize endpoints + reassembly | 2.5h |
| UI: progress bar + "Connection lost — retrying" + cancel | 1.5h |
| Tests + manual cellular smoke | 1.5h |

**Total: ~8h.**

jayzalowitz:
[...]

PR #460, opened 2026-05-31
p2.6: resumable chunked voice upload backbone (#386)

jayzalowitz:
## Summary

Closes the resilient-upload backbone of #386 P2.6. A 2-minute voice memo (~3MB base64) went up as a single POST; a cellular drop mid-upload meant starting over with no progress and no retry. This adds a resumable chunked path end-to-end at the service layer.

### API
- **`apps/api/src/lib/upload-session.ts` (new)** — `UploadSessionStore`: in-memory, TTL-swept (10min idle), ownership-checked. Accepts base64 string chunks in any order, replaces retried chunks in place (no double-count vs the size cap), reports the missing-index list, concatenates in index order on finalize (decoded once → no per-chunk alignment concern).
- **4 endpoints** — `POST /api/voice/upload/{session,chunk,finalize,cancel}`, sharing the whisper tail with `/transcribe` (refactored into a `runTranscription` helper). Inherits `sessionAuth + requireOwnership` (body `userId`) + the store's per-session owner check. 25MB ceiling mirrored from the single-shot path.

### Mobile
- **`voice-chunker.ts` (new, pure)** — `chunkBase64` / `countChunks` / `reassembleChunks`.
- **4 `voiceUpload*` methods** on `SkyTwinApiClient`.
- **`transcribeRecordingChunked`** — opens a session, uploads each chunk with **per-chunk** exponential-backoff retry (1s→8s cap, 4 attempts), `onProgress` with a `retrying` flag, cooperative `isCancelled()` cancellation, finalize.

## Acceptance criteria

- [x] Disconnect mid-upload → resume from the next chunk after reconnect (per-chunk retry; server keeps received chunks; 409 lists the missing index)
- [x] Cancel mid-upload (cooperative `isCancelled()` → server cancel)
- [ ] **Visible progress bar in the UI** — the `onProgress`/`isCancelled` hooks are exposed and tested; the `VoiceScreen` visual wiring (progress bar + "Connection lost — retrying" + cancel button) is a **follow-up on #386** (needs the live API + a device to verify)

## Test plan

- [x] `pnpm --filter @skytwin/api test` — 779 pass (13 store + 4 route integration new)
- [x] `pnpm --filter @skytwin/mobile test` — 212 pass (12 chunker + 6 orchestrator new)
- [x] API `tsc --noEmit` clean; mobile unchanged at the 6 pre-existing base-`main` errors
- [ ] Manual cellular airplane-mode-mid-upload smoke — deferred to a device pass

## Scope note

Backbone (server + pure logic + service orchestration) now; the `VoiceScreen` visual progress UI is a follow-up on #386. Mirrors the slicing used on #387 (routing now, native actions later).


jayzalowitz:
Addressed all 3 findings in a7cfc1d:

1. **Body limit (real blocker)** — the global `express.json()` 100KB default would 413 a ~256KB chunk before the router ran (and was already 413-ing `/transcribe`'s up-to-25MB payloads in production). Added a voice-scoped `express.json({ limit: '30mb' })` **before** the global parser; it parses `/api/voice` bodies first so the global one skips them, while non-voice routes keep the tight 100KB default.
2. **Whole-payload base64 validation (real correctness)** — per-chunk validation let a non-final chunk end in `=` padding; once concatenated, Node's decoder silently truncates after the interior `=`. `runTranscription` now re-validates the reassembled string (regex + `length % 4`) before decode and 400s on failure. This also caught that my integration test built **unrealistic** chunks (separately base64-encoded pieces) — fixed to slice ONE valid base64 string the way `voice-chunker` does, plus a new test that an interior-padding reassembly is rejected with no transcribe call.
3. **Cancel during backoff** — `isCancelled()` was only checked before each chunk, so cancelling mid-backoff waited the full delay and burned one more retry. Added a post-sleep cancel check that aborts + cancels the server session. New test covers it.

780 API + 213 mobile tests pass; API tsc clean.

PR #447, opened 2026-05-26
p1.5: LLM provider chain is local-first by default (#375 slice)

jayzalowitz:
## Summary

Pre-fix, the default LLM provider chain was \`ANTHROPIC → OPENAI → GOOGLE → OLLAMA → EMBEDDED\`. **The moment a single cloud key was set in the env, every decision's prompt left the machine** — sender, subject, snippet, preferences, patterns, traits, last 5 episodic memories. The privacy policy promised \"your data stays local\"; the default contradicted that the instant a user configured Anthropic for \"smarter decisions.\"

This PR ships the **default-order fix**. The per-user toggle UI and the prompt-builder PII redaction pass — the other halves of the original #375 spec — are tracked as follow-ups; closing the doc-vs-behaviour gap is the most acute leak and a single env-var-aware reorder addresses it without a UI change.

## What's in this PR

| Change | Effect |
|---|---|
| \`buildProviderChain\` splits providers into local (embedded + ollama) and cloud (anthropic + openai + google) buckets | Local providers always come first by default |
| New \`SKYTWIN_LLM_PRIORITY=cloud-first\` env var | Opt-in path back to legacy hosted-first ordering — explicit consent gate, env var is the audit trail |
| Unknown / typo values fall through to local-first | A typo must NOT silently escalate to cloud |
| \`buildProviderChain\` exported (was private) | Ordering invariants are unit-testable |
| \`docs/privacy.html\` updated | Documents the new ordering + the opt-in env var |

## Tests

8 new in \`apps/api/src/__tests__/llm-client-factory.test.ts\`:
- Default → local-first order
- \`SKYTWIN_LLM_PRIORITY=cloud-first\` → legacy order
- Typo (\`clouf-first\`) → falls back to local-first
- Case-insensitivity (\`CLOUD-FIRST\`)
- Empty value → local-first (the default)
- Empty config → empty chain regardless of priority
- Only-cloud → canonical cloud sub-order stable
- Only-local → canonical local sub-order stable

## Follow-ups (same issue)

- Per-user toggle UI (Settings → AI provider + first-run wizard step)
- Prompt-builder PII redaction pass (replace raw event JSON with a redacted view before sending to cloud)

## Test plan
- [x] \`pnpm --filter @skytwin/api build\` — clean
- [x] \`pnpm --filter @skytwin/api exec vitest run src/__tests__/llm-client-factory.test.ts\` — 18 pass (10 existing + 8 new)
- [ ] CI matrix
[...]

Issue #375, opened 2026-05-25
P1.5 LLM prompts include user memory and email senders with no redaction


jayzalowitz:
**Parent:** #361

## Context

SkyTwin's privacy story has two surfaces: data-at-rest (covered by #374) and data-in-flight to LLM providers. This issue covers the in-flight side.

The privacy doc and the homepage both say variants of "your data stays local." A user reads that, configures an Anthropic API key in Settings so the twin's reasoning gets sharper, and assumes the local-first promise still holds. It does not. Every situation interpretation and every candidate generation sends `rawEvent` (full email metadata including sender, subject, snippet), the user's recent preferences, behavioral patterns, cross-domain traits, and the last five episodic memories to whichever provider is in the chain. With the default chain ordering — Anthropic → OpenAI → Google → Ollama → embedded — the moment one cloud key is present, the cloud provider becomes first-call and the user's PII leaves the machine on every decision.

The launch-day stranger reading "Your Google data never leaves your machine," then enabling Anthropic for "smarter decisions," is being surprised — and they will rightfully feel deceived.

## Root Cause Analysis

Three structural facts compound:

1. **The prompt builder sends the raw event verbatim.** `packages/llm-client/src/prompt-builder.ts:12-35` calls `JSON.stringify(rawEvent, null, 2)` and pastes it into the situation-classification prompt. For email signals, `rawEvent` includes the sender's email address, the subject line, the body snippet, the thread participants, and any labels. The candidate prompt at lines 41-98 layers on `decision.rawData` plus preferences, patterns, traits, and episodic memories. There is no redaction pass.

2. **The default provider chain puts cloud providers first.** `apps/api/src/lib/llm-client-factory.ts:27-78` appends providers in order: Anthropic (if `ANTHROPIC_API_KEY` set), then OpenAI, then Google, then Ollama, then embedded. The chain is "first-success-wins"; if Anthropic is configured it is always called first. A user who configures any cloud key has just made cloud-first the default, with no in-product confirmation step that this was an informed choice.

3. **The privacy policy does not name the choice.** The doc says "your data stays local" without "unless you configure a cloud LLM." A reader cannot reconcile the promise with the behavior because the conditional was never written down.

The intent of the prompt builder is correct: give the LLM enough context to reason well. The intent of the provider chain is correct: prefer the highest-quality model the user has authorized. Both intents are right; the **default ordering** and the **policy text** are what break the local-first promise.

## Verified Current State

[...]

PR #462, opened 2026-06-01
docs(p1.5): privacy summary names the cloud-AI opt-in exception (#375 fix c)

jayzalowitz:
## Summary

Closes the launch-scoped documentation half of #375 P1.5. `docs/privacy.html`'s top-line **Summary** said *"Your Google data … never leaves your machine"* unconditionally — contradicting (a) the detailed "Data sharing & transfers" section right below it, which already documents the opt-in cloud-AI path, and (b) the exact behavior #375 flags: a user reads the summary, enables Anthropic in Settings for sharper reasoning, and is surprised their email senders/snippets now leave the device.

- Summary now states the local-first default explicitly (built-in on-device model → nothing leaves) **and** names the one opt-in exception (enabling a cloud AI provider sends each decision's content to that provider), linking to a new `#ai-providers` anchor on the detail section.

## Why this matters for launch

- **#375 fix (c)** — the launch-scoped doc fix. (fix (a) local-first chain default shipped in #447; fix (b) in-prompt PII redactor is the issue's explicitly-deferred v0.8 item.)
- **#351 Google OAuth restricted-scope review** — reviewers read this exact summary and verify stated data flows match behavior. An unconditional "never leaves" that the detail section contradicts is the kind of inconsistency that fails review.

## Test plan

- [x] Documentation-only; no code change. The `#ai-providers` anchor target was added to the "Data sharing & transfers" heading so the summary's in-page link resolves.


jayzalowitz:
Addressed in 2458330: reworded the opening line so it no longer flatly claims "no third party" — it now says no transfer **for the developers' benefit** (no analytics/ads/training/broker pipeline, no SkyTwin server receives it) and explicitly flags item 2 (the opt-in cloud AI provider you enable) as the single third-party case. The section now reads consistently with both the list item and the Summary.

PR #468, opened 2026-06-02
fix(db): remove dead acquireTimeoutMillis config (#378 follow-up)

jayzalowitz:
## Summary

Follow-up to #378. That PR added `acquireTimeoutMillis: 5000` to `DatabaseConfig` and `DEFAULT_CONFIG`, but:

1. It was **never passed into `new Pool({...})`** in `getPool()` — only `connectionTimeoutMillis` reaches the Pool.
2. **pg-pool never reads it.** Verified against the resolved `pg-pool@3.14.0` source (`pg` 8.21.0 on main): the only timeout key it honors is `connectionTimeoutMillis`. `acquireTimeoutMillis` isn't in `@types/pg`'s `PoolConfig` either.
3. The field's JSDoc claimed *"Supported in pg-pool ≥ v3.6"* — **false.**

Copilot flagged this on #378 ("won't actually enforce an acquire timeout... Changelog/inline comments will be misleading"); it merged before the comment was addressed.

## The behavior #378 wanted is already correct

`connectionTimeoutMillis: 5000` (already wired) **is** the acquire-from-saturated-pool bound. pg-pool uses that one value for both new-connection establishment and the wait for a free slot when the pool is full:

> `pg-pool/index.js` — when `_isFull()`, the pending-queue waiter is armed with a `setTimeout(..., this.options.connectionTimeoutMillis)` ("set connection timeout on checking out an existing client").

So the 21st concurrent acquire on a `max: 20` pool already fails within ~5s with a clear timeout instead of hanging. **No behavioral change in this PR** — the acquire timeout was, and remains, 5s.

## Change

- Remove the dead `acquireTimeoutMillis` field from the `DatabaseConfig` interface and `DEFAULT_CONFIG`.
- Correct the inline comment to attribute the saturated-pool bound to `connectionTimeoutMillis` (with the pg-pool source reference).
- Fix the #378 CHANGELOG entry's `(1)` clause to match reality.

## Verification (local)

- `pnpm build` — 35/35 tasks
- `pnpm test` — 70/70 tasks
- Confirmed `acquireTimeoutMillis` no longer appears in `packages/db/dist/connection.d.ts` or anywhere in source.


jayzalowitz:
## /review — CLEAN (0 findings)

Dead-code removal + comment correction; no behavioral change.

**Verified:**
- No consumer constructs `DatabaseConfig` with `acquireTimeoutMillis` (grep: only the stale `dist/` artifact, regenerated clean). `pnpm build` 35/35 confirms the interface change broke nothing.
- Behavior preserved: `connectionTimeoutMillis: 5000` is the real acquire bound, confirmed against pg-pool 3.14.0 (`_isFull()` arms the pending-queue waiter with it). Acquire timeout stays 5s.
- Kept the accurate `connectionTimeoutMillis` JSDoc that already documented the saturated-pool behavior; removed only the false `acquireTimeoutMillis` one.
- `pnpm test` 70/70 local.

Issue #380, opened 2026-05-25
P1.10 README version badge is 35 releases stale


jayzalowitz:
**Parent:** #361

## Context

The README is the first impression for every cold-loaded visitor — HN reader, GitHub-Trending lurker, friend forwarded a link, recruiter cross-referencing a resume. The version badge at the top says `v0.6.23.2`. The Project Status paragraph says `(v0.6.23.2)` and "active development." The actual current version in `package.json` is `0.6.58.0`. The CHANGELOG has 126 `## [` heading entries since (35+ minor releases inside the 0.6.x series). A visitor reading "0.6.23.2, active development" sees a project that is either sleepy or recently-abandoned.

This is a two-line text bug. It is also the bug that has the highest leverage-per-unit-effort of any item in Epic D, because it shapes the trust a stranger forms in the first 10 seconds before they read anything else. A README that lies about its own version reads as a README nobody maintains, which reads as a codebase nobody maintains.

Adjacent gap: the README has no monetization line. The launch plan and CHANGELOG both reference a free-personal / paid-team split, but a stranger has no way to tell whether they're looking at a hobby project, a venture-funded teaser, or something with a sustainable model. One sentence fixes it.

## Verified Current State

Verified 2026-05-25 against `jayzalowitz/launch-skytwin-demo @ 8bb5379`.

| File | Lines | Current value |
|---|---|---|
| `README.md` | 9 | `<img src="https://img.shields.io/badge/version-0.6.23.2-green.svg" alt="Version">` |
| `README.md` | 372-374 | "SkyTwin is in **active development** (v0.6.23.2). The core decision pipeline, twin model, policy engine, and swappable memory layer are functional. ..." |
| `package.json` | (root) | `"version": "0.6.58.0"` |
| `CHANGELOG.md` | (full) | 126 `## [...] - YYYY-MM-DD` entries, most recent: `## [0.6.58.0] - 2026-05-23`, then `## [Unreleased] — Tier 2 launch polish`, then `## [0.6.57.0] - 2026-05-22`, ... |
| `docs/launch-plan.md` | 1-9 | "Tier 0 — What's already shipped (in PR #350)" reflects the bundled-install / Gmail-wizard / Tier-0 status. Project status in README should reflect this tier (post-Tier-0, in Tier 1) — not the generic "active development" claim. |
| `.github/workflows/` | (3 files) | `build.yml`, `evals.yml`, `install-validation.yml` — no release-prep step that writes the version badge. |
[...]

Issue #370, opened 2026-05-25
P0.9 Auto-update feed URL is a non-existent domain


jayzalowitz:
**Parent:** #359

## Context

Once we ship a signed `.dmg` (#368), the next question is "how do users get the next version?" Today, the answer is "they don't." `apps/desktop/src/auto-update.ts` defaults the update feed URL to `https://updates.skytwin.local/` — a non-existent domain. `electron-updater` will get a DNS resolution failure on every `checkForUpdates()` call and silently log an error.

This means: a security bug in v0.7.0 is permanent for every installed user. A policy-engine fix in v0.7.1 reaches nobody. The auto-update path that already exists in code (`AutoUpdateController` at `auto-update.ts:127`, wired into `main.ts:133`) is plumbed but pointed at nothing.

The good news: the code is already set up for GitHub Releases as a publisher. `package.json` has `"provider": "github"` in the build.publish config (lines 101-106), and `ElectronUpdaterBackend` (`auto-update.ts:53-80`) uses `electron-updater`'s built-in GitHub provider. So Option A — use GitHub Releases as the update feed — is approximately one config change and a passes-all-tests away.

The bad news: the `.local` fallback in `resolveFeedURL()` (`auto-update.ts:114`) is the active default in CI today because `SKYTWIN_UPDATE_URL` is set nowhere. Verified: `grep -r SKYTWIN_UPDATE_URL .github/` returns zero CI/secret references; only the `CHANGELOG.md` history mention.

## Root Cause Analysis

The auto-update plumbing has three layers:

1. **Backend selection** (`auto-update.ts:90-98`): `defaultBackend()` returns `ElectronUpdaterBackend` inside a real Electron process and `NoopUpdateBackend` everywhere else. This is correct.

2. **Configuration** (`auto-update.ts:110-115, 118-125`): `resolveFeedURL()` returns the SKYTWIN_UPDATE_URL env var if set, otherwise `https://updates.skytwin.local/`. `defaultAutoUpdateConfig()` builds the config from these.

3. **electron-updater publisher** (`apps/desktop/package.json:101-106`): The `build.publish` block sets `provider: github`, `owner: jayzalowitz`, `repo: skytwin`. electron-updater reads this at runtime to know WHERE to fetch updates from.

[...]

Issue #368, opened 2026-05-25
P0.7 Desktop builds are unsigned; Gatekeeper-blocked


jayzalowitz:
**Parent:** #359

## Context

Every installer SkyTwin has ever produced is unsigned. macOS Gatekeeper shows "SkyTwin can't be opened because Apple cannot check it for malicious software." Windows SmartScreen shows "Microsoft Defender SmartScreen prevented an unrecognized app from starting." A friend who has been told by IT to be careful about unsigned binaries will not click through. A real stranger downloading from a tweeted GitHub Releases link will not even try.

This is the launch-blocker that no engineering work alone can resolve — it requires purchasing certificates, waiting for issuance, and plumbing six secrets into CI. The code path for signing already exists (`apps/desktop/scripts/sign-and-notarize.ts`). The certificates and the secret wiring do not.

## Root Cause Analysis

Three coordinated absences:

1. **No certificates have been acquired.** Apple Developer Program ($99/yr, ~1-5 day approval) and Windows EV code-signing certificate ($300-600/yr, 3-7 day identity validation) are both prerequisites. Neither has been purchased.

2. **CI explicitly disables signing.** `.github/workflows/build.yml:127` sets `CSC_IDENTITY_AUTO_DISCOVERY: 'false'` on the macOS packaging step. This was added to prevent electron-builder from probing the runner's keychain for an identity that does not exist (which would fail the build with an unhelpful error). The flag is correct given the current state — there is no identity to discover — but it is the codified "do not sign" decision.

3. **`release.yml` has been deleted.** Recent commits (`8bb5379` revert to softprops, `feafd2f` skip signing when no cert secrets, ending in `release.yml` deletion in #356) mean the only release path is `build.yml`'s `release:` job (lines 430-464), which downloads pre-built unsigned artifacts and attaches them to a draft GitHub Release via `softprops/action-gh-release@v3`. There is no signing step anywhere in the release path.

The infrastructure that exists today — `package:signed` script (`apps/desktop/package.json:22`) and `sign-and-notarize.ts` validation wrapper — is dormant. `SKYTWIN_SIGN_RELEASE=true` is never set anywhere in CI. Build is wired to unsigned and stays unsigned.

## Verified Current State

Verified 2026-05-25 against `jayzalowitz/launch-skytwin-demo @ 8bb5379`.

| File | Lines | What it does today |
|---|---|---|
[...]

PR #518, opened 2026-06-15
feat(ci): ship electron-updater manifests + feed-reachability gate (#370)

jayzalowitz:
## What & why

Closes the remaining **code** half of #370 (the launch-readiness audit's scope hint). The desktop auto-update feed is GitHub Releases via electron-updater's `github` provider (`apps/desktop/package.json#build.publish`). An installed app's `checkForUpdates()` reads `latest-mac.yml` (macOS) / `latest.yml` (Windows) / `latest-linux.yml` (Linux) off the latest Release to discover a new version. Pre-fix, the `release` job in `.github/workflows/build.yml` attached only the installer binaries — those manifests were absent, so **every update check failed**: a fix in vN never reached an installed vN-1.

electron-builder writes the manifests into `dist-electron/` during packaging **even under `--publish never`** (that flag only suppresses the *upload*, not the *generation*), so the fix is purely collection-and-attach — no re-architecture of the publisher path, no change to `--publish never` semantics, no change to the softprops draft-release job.

## Changes

- **`.github/workflows/build.yml`**
  - desktop-mac / desktop-windows: upload `dist-electron/latest-mac.yml` / `latest.yml` as artifacts (`if-no-files-found: error`).
  - desktop-linux: upload `dist-electron/latest-linux.yml` (`if-no-files-found: warn` + `if: always()`, matching the Linux package step's existing `continue-on-error: true` tolerance — a Linux packaging hiccup must not fail the whole tagged build).
  - `release` job: list the three manifest artifacts in `files:`.
  - `release` job: new first step **Verify update feed reachable** — `curl -fsSL --max-time 10 --retry 3 "https://github.com/<repo>/releases/latest"`, fails the job on any non-2xx.
  - **Unchanged:** all three `--publish never` package steps; softprops draft-release config.
- **`apps/desktop/src/__tests__/release-update-manifests.test.ts`** (new) — 7 vitest guard cases pinning the manifest globs, the release-attachment list, the preserved `--publish never` count, and the curl gate (mirrors the existing "README badge must stay dynamic (#380)" string-guard pattern).
- **Docs** — `docs/launch-plan.md` §2.1 + `docs/release-procedure.md` gap 2 updated from "manifests not generated" to "manifests ship; live end-to-end only after signing". CHANGELOG `[Unreleased]` entry added.

## Acceptance criteria

- **AC#2** (CI verifies the GitHub Releases endpoint is reachable before the `release` job publishes; fails on non-2xx) — ✅ `Verify update feed reachable` step.
- **AC#6** (`latest-mac.yml` / `latest.yml` / `latest-linux.yml` attached to every GitHub Release) — ✅ artifact-collect + `files:` attach.
- ACs #1, #3, #5, #7 (the `auto-update.ts` `.local` removal + test update) — already shipped in PR #453; unchanged here.

## Tests

`pnpm --filter skytwin-desktop test` → the new guard file (7 tests) and the existing `auto-update.test.ts` (31 tests) pass. The 3 failures in `integration-live.test.ts` are pre-existing/environmental (they hit a live API at `localhost:3100` that isn't running in CI sandbox); untouched by this PR.

## Scope note

The self-update path is live end-to-end only after **code signing (#368/#359)** — electron-updater refuses an unsigned update payload (fails safe). The manifests shipping now is harmless and unblocks the fresh-box self-update verification (#370 AC#3/#4) once signed binaries land. Hence **Addresses #370**, not Closes.


jayzalowitz:
**Serial-merge pipeline (2026-06-15): deferred — needs a tagged-build verification before landing (release-CI blast radius).**

The diff is clean and well-commented (additive manifest-upload steps + a curl feed-reachability gate + 3 `files:` lines on the `release` job), and rebases without conflict. But it rests on one assumption I can't verify in-repo: that **electron-builder writes `latest-mac.yml` / `latest.yml` / `latest-linux.yml` into `apps/desktop/dist-electron/` even under `--publish never`**. electron-builder ties `latest*.yml` generation to its publish step; with `--publish never` it may NOT emit them. The mac/win upload steps use `if-no-files-found: error`, so if that assumption is wrong the **desktop jobs fail on every tag push** (and on any PR that runs them), breaking the release pipeline.

To land safely: run a real `pnpm --filter skytwin-desktop run package:mac --publish never` and confirm `dist-electron/latest-mac.yml` exists. If it does → merge as-is. If it does NOT → the manifests need an explicit generation path (e.g. `electron-builder --publish onTagOrDraft` with `GH_TOKEN` gated to tags, or `--config.publish` to write-without-upload), not the current `--publish never`. Either way this is a tag dry-run gate, not a code-review call. Everything else (#370's dead-config half) already shipped in #453; this is the manifest-generation half. Leaving open with the verification step recorded.

Issue #401, opened 2026-05-25
P3.3 OS-keychain integration for credential-vault passphrase


jayzalowitz:
**Parent:** #357

Currently users re-type the passphrase on every restart. Add macOS Keychain / Windows Credential Manager / Linux Secret Service integration as an optional convenience.

### Acceptance criteria
- [ ] First unlock → "Remember on this device?" prompt.
- [ ] If yes, passphrase stored in OS keychain.
- [ ] If no, current behavior unchanged.

Issue #374, opened 2026-05-25
P1.4 User memory and preferences are stored unencrypted


jayzalowitz:
**Parent:** #361

## Context

SkyTwin's privacy promise is "your data stays local." The homepage, the onboarding wizard, and the safety doc all repeat a variant of "Your Google data never leaves your machine." Technically true on the wire: signals and decisions live in the embedded CockroachDB on the user's disk, never sent to a cloud. But the at-rest reality does not match the promise.

A stolen laptop, a leaked backup, a piece of credential-stealing malware, or any tool with read access to the CRDB data directory exposes every user fact, every memory, every relationship the twin learned — in plaintext. The user reads "stays local" and infers "private even if my laptop is stolen." Today that inference is wrong.

For the launch-day stranger trusting the twin with their inbox, their schedule, and the relationships extracted from years of correspondence, this is the privacy gap they cannot personally detect. There is no UI surface that says "by the way, all of this is plaintext." There is only the privacy doc saying "stays local."

The fix already exists in seed form. `packages/credential-vault/` shipped envelope encryption for OAuth tokens in migration 032 (#212). Every primitive needed to extend it — scrypt KDF, AES-256-GCM, per-row IV, key cache, key rotation — is on disk. The work is to widen the boundary from "tokens only" to "tokens + memory + preferences + episodic + brain."

## Root Cause Analysis

Three structural facts compound:

1. **Envelope encryption was scoped to OAuth tokens only.** Migration `032-encrypted-oauth-tokens.sql` adds encrypted columns to `oauth_tokens` and the per-user vault metadata (`user_credential_vault_meta`). Every other table that stores user-derived content was left plaintext. Searching `packages/db/src/repositories/*.ts` for envelope / `encryptedAccess` / `encryption_key_version` / `deriveKey` returns one match: `oauth-repository.ts`. The rest of the data layer never learned about the vault.

2. **The most sensitive tables are the ones with no encryption.** `brain_pages.content` is the entire textual surface of the brain backend — extracted facts, summaries, code definitions. `memory_drawers.content` holds the legacy MemPalace's atomic memory chunks. `episodic_memories.situation_summary` plus its `context_snapshot` and `outcome` JSONB hold "what happened, with whom, on what day, with what feeling." `preferences.value` holds the twin's learned preferences. All four are the things a privacy-aware user would expect to be the most-protected, and all four are stored in CRDB as readable strings.

[...]

PR #520, opened 2026-06-15
feat(db): at-rest envelope encryption for high-value tables (#374)

jayzalowitz:
## What & why

SkyTwin's privacy promise is "your data stays local," but until now everything except OAuth tokens sat in the local CockroachDB as **plaintext** — a stolen laptop or imaged disk exposed every learned preference, twin-profile inference, and extracted memory. CockroachDB Core (what the desktop ships) has no TDE, so the fix has to live in the application.

This PR extends the existing `@skytwin/credential-vault` envelope (AES-256-GCM + scrypt-derived per-user key — the same machinery that already protects OAuth tokens via migration 032) to the **three highest-value tables**, scoped strictly per the launch-readiness audit:

| Table | Encrypted columns |
|---|---|
| `preferences` | `value`, `evidence` |
| `twin_profiles` | all 7 JSONB columns |
| `brain_pages` | `title`, `content`, `metadata` |

## Changes

- **Migration `065-encrypt-high-value-tables.sql`** — adds `<col>_encrypted BYTES` + `encryption_key_version INT` to all three tables and makes the plaintext columns nullable, mirroring migration 032's lazy-migration shape. Plaintext columns are kept for the migration window; a **separate follow-up PR drops them after a soak** (per the issue's rollback plan).
- **`packages/db/src/lib/vault-helper.ts`** (new) — shared `encryptColumn` / `decryptColumn` (packed `IV(12)+tag(16)+ciphertext`, fresh IV per call, identical to the `DbTokenStore` precedent), `readColumn` (the lazy-migration decision table), `resolveKey` (`plaintext` when off / `unlocked` / `locked`). All expected failures are **typed result objects, never thrown**; a locked vault returns `vault_locked` and **never leaks ciphertext**.
- **`preferences` wired end-to-end** through `TwinRepositoryAdapter`: encrypt-on-write (NULLs the plaintext sibling), decrypt-on-read, plaintext fallback during the migration window, `VaultLockedError` surfaced on locked read/write (for the API to map to `409 vault_locked`). Opt-in per process via `setPreferenceVaultKeyProvider(keyCache)`; **no-op plaintext (fully backward compatible) when unwired**.
- **Privacy policy** (`docs/privacy.html`) — new "Memory and preferences are encrypted at rest" subsection documenting what's encrypted, the passphrase-derived key, and the deliberate v0.7 trade-off (FTS `content_tsv` + `embedding` stay plaintext for search performance).
- **CHANGELOG** entry under `[Unreleased]`.

Master key derives from the existing credential-vault passphrase mechanism; OS-keychain integration is the tracked follow-up (#401).

## Acceptance criteria

- **AC1** (migration columns for the scoped tables): ✅ all 3.
- **AC2** (encrypt-on-write / decrypt-on-read / plaintext fallback): ✅ `preferences`.
- **AC4** (ciphertext unreadable as STRING; round-trips with the right key): ✅ unit tests.
- **AC5** (vault-locked read never returns ciphertext — returns `vault_locked`): ✅ unit tests.
- **AC7** (privacy policy documents at-rest + tsv/embedding trade-off): ✅.
- **AC9** (existing OAuth-token tests pass unmodified): ✅ (9/9, plus 18/18 twin-profile).

## Tests

26 new vitest cases (`vault-helper.test.ts` ×19, `preferences-vault.test.ts` ×7): round-trip per data shape (STRING / JSONB / array), ciphertext-unreadable-as-string, fresh-IV, wrong-key / tampered / malformed → `decrypt_failed`, vault-locked → `vault_locked` (short-circuits before querying the DB), lazy-migration plaintext fallback, feature-off plaintext path. **Full `@skytwin/db` suite: 363 passed, 19 skipped (E2E-gated), 0 failed.** `tsc` on the package is clean for all changed files (remaining errors are pre-existing unbuilt-transitive-dep artifacts in a fresh worktree, unrelated to this change; CI builds the full graph).

## Documented remainder (status: partial — tracked under #374)

This ships migration columns for all 3 scoped tables + full repo wiring for `preferences`. Still open:
- `twin_profiles` + `brain_pages` read/write wiring (the latter coordinated with the `brain_pages.metadata->>'authoringTier'` tier-weighting SQL in `@skytwin/memory-gbrain-crdb-adapter`).
- Lazy-migration backfill worker (AC3), `409 vault_locked` API middleware, Settings vault-status + lock-button UI (AC8), psql/perf integration tests (AC6, AC10).
- The remaining ~11 tables from the issue's audit (decisions, signals, explanation_records, memory_*, knowledge_*, episodic_memories, assistant_threads, lifebooks, …).

Addresses #374


jayzalowitz:
**Serial-merge pipeline (2026-06-15): deferred — needs the key-management decision + a migration renumber, not a mechanical merge.**

Two blockers, neither a rebase:
1. **Key-management decision (the real one).** #374 encrypts memory/preferences at rest with envelope encryption, but the master-key strategy is a deliberate security choice (passphrase-derived vs OS-keychain-backed — now that #401's `safeStorage` keychain landed via #514, that's a live option). Shipping at-rest encryption with the wrong key model is worse than shipping none. This needs a human call, consistent with the launch-readiness report's #1 code-side item.
2. **Migration number collision.** This PR adds `065-encrypt-high-value-tables.sql`, but `065-worker-dead-letter.sql` already landed on `main` (via #506). The migration must be renumbered to the next free slot (066) and any references updated.

Recommend: settle the key-management approach (link to #401), renumber the migration, re-run the suite, and land in a dedicated reviewed PR. This is the top engineering pre-launch item; it should not be force-merged at the tail of a batch. Leaving open.

PR #555, opened 2026-06-25
v0.6.71.0 fix(dev): supervise openclaw bridge

jayzalowitz:
## Summary

- add a dev-only supervisor for `@skytwin/openclaw-bridge` so a one-off child `SIGKILL` / exit 137 restarts the bridge instead of letting Turbo tear down API/web/worker
- keep production `start` on the direct `node server.mjs` path
- add a node:test regression that starts the supervisor, kills the child bridge process with `SIGKILL`, and asserts a replacement child starts
- bump release metadata to `0.6.71.0` and document the dev behavior

## Root Cause

The log showed `@skytwin/openclaw-bridge#dev` exiting with code 137 after a long-running `pnpm dev`; Turbo then sent termination signals to the other persistent tasks. The bridge process is optional dev infrastructure, but it was run directly as a persistent Turbo task, so any external kill of that process became a full-stack dev shutdown.

## Verification

- `pnpm --filter @skytwin/openclaw-bridge test`
- `pnpm build --concurrency=1`
- `pnpm test`
- `pnpm lint`
- manual reproduction: ran `pnpm --filter @skytwin/openclaw-bridge dev`, verified `/health`, killed the child `server.mjs` with `SIGKILL`, verified a new child PID and `/health` recovered


jayzalowitz:
Addressed both Copilot comments from the first review pass:\n\n- Restart tuning env vars now go through `readIntegerEnv(...)`, which rejects invalid/non-integer/out-of-range values, logs a warning, and falls back to safe defaults. Added regression test: `dev supervisor falls back to safe defaults for invalid restart config`.\n- `stop()` now waits for the child and escalates to `SIGKILL` after `OPENCLAW_BRIDGE_STOP_GRACE_MS` if the child ignores shutdown. Added regression test: `dev supervisor SIGKILLs a child that ignores shutdown`.\n\nValidation after the fixes: `pnpm --filter @skytwin/openclaw-bridge test`, `pnpm build --concurrency=1`, `pnpm test`, and `pnpm lint` all pass locally.

PR #558, opened 2026-06-25
fix(decision-engine,api): coverage panel reads real accounts + understands alternative providers

jayzalowitz:
## What

After the Outlook connectors landed (#557), the digest's **"what I can see" coverage panel went all-red for a Microsoft-only user** — every capability "unavailable", with prompts to connect Gmail/Google Calendar — even though their Outlook mail + calendar were actively producing commitments. The inverse was latent too: a Google user would have been nudged to *also* connect Outlook.

Root cause: `computeCoverage` treated every source in a capability's allow-list as separately required (`available = all connected`), which has no notion of **alternative** providers.

## Fix

Compute coverage over **source equivalence groups**:
- `gmail` / `outlook` → the **email** group
- `google_calendar` / `outlook_calendar` → the **a calendar** group
- `filesystem` → files, `voice` → voice

A group is satisfied by **any** connected member. So:
- A Microsoft-only user gets the **exact same capability statuses** a Google user gets (asserted by a parity test).
- Neither vendor's user is nudged to connect the other.
- `unlockedBy` / `missing` now carry **human-meaningful group labels** ("connect a calendar") instead of raw, redundant source ids ("connect google_calendar, outlook_calendar").
- `normalizeProvider` maps `microsoft` → `[outlook, outlook_calendar]`.

The `security` matrix set is unchanged — `gmail`'s membership already puts the email group in scope, and `outlook` is a member of that group, so Outlook mail correctly enables security coverage without editing the matrix.

This **closes the deferred limitation** documented in #557.

## Tests

11 unit tests in `source-coverage.test.ts`, including Microsoft↔Google status parity, the no-cross-vendor-nudge guarantees, and group-label assertions. decision-engine 324 / api 889 green.


jayzalowitz:
## /review + adversarial pass

Ran the merge-gate `/review` (critical pass + independent adversarial subagent) on this diff. The adversarial pass **confirmed the group-coverage logic correct on all combinations** (traced []/gmail/google_calendar/filesystem/voice/microsoft/google/gmail+outlook/gmail+google_calendar/microsoft+voice — no over- or under-crediting; security-on-Outlook is semantically right; determinism + coldStart sound; the label change breaks no consumer — `twin-briefing.js` only displays them).

It also caught a **P1 that made this PR a no-op in production**, now fixed in `e62a1b1`:

**Coverage read a dead table.** `computeCoverage` was fed from `connected_accounts`, which **nothing writes to** — OAuth persists to `oauth_tokens`. So every real user got zero accounts → `coldStart:true`, and the "what I can see" panel was stuck in cold-start for *everyone* (Gmail users included), which also meant the alternative-provider fix would never have been visible. Repointed the query at `oauth_tokens` (the OAuth source of truth — a row exists iff the connection is live; `refresh_token` not filtered since it's NULL for vault-migrated rows). Added two regression tests asserting a Google connection yields a non-cold-start panel with security available, and a Microsoft connection yields identical statuses with no nudge to connect Google.

Scope of the PR is now: **make the coverage panel correct** — read real connected accounts *and* treat Gmail/Outlook (and Google/Outlook calendars) as alternative providers. Suites: decision-engine 324, api 891, all green.

PR #565, opened 2026-06-26
feat: awareness disposition gate — stop the approval-queue flood (flagged, default off)

jayzalowitz:
## What — Root 2 from the newsletter audit (the structural root)

#562/#564 fixed *which action* each signal gets. This fixes *how many cards*. At `observer`/`suggest` tier, `policy-evaluator.ts:464` forces `requiresApproval` on **every** action and `events.ts:521` creates one approval card per signal — so newsletters, automated notices, the user's own re-ingested sent mail, and "no action required" calendar updates each still become a "Needs your OK." The DESIGN.md act-vs-aware split + digest are **read-time only**; they display the flood, they don't stop it.

## The gate

A write-path disposition gate. `isAwarenessOnly(decision, outcome)` is true when:
- the selected action is **passive + reversible + zero-cost** (`archive_email` / `label_email` / `create_note` / `acknowledge` / `dismiss`), **and**
- the context is awareness — a `CALENDAR_UPDATE`, or an `EMAIL_TRIAGE` whose `authoringTier` ∈ `{inbox_newsletter, inbox_automated, user_sent_originated, user_sent_reply}`.

When gated, the decision/outcome/explanation are still persisted (the digest surfaces it as FYI) but **no approval row and no `approval:new` SSE**. `selectedAction` stays non-null, so `decision:blocked-by-policy` also stays silent.

## Hard safety carve-outs (never gated)
- **Injection-guard escalations** — `outcome.confirmationLevel` set ⇒ a security escalation on untrusted content ⇒ always surfaces (`policy-evaluator.ts:105` never strips this).
- **Human inbound mail** — `inbox_personal` / `inbox_broadcast` are deliberately *excluded*; a cc'd human thread can be important.
- **Calendar invites**, and any non-passive / irreversible / costed action.

## Flagged + phased
- **Phase 0 (this PR, always on):** logs every awareness candidate with **no behaviour change** — so production logs confirm it fires on newsletters/sent-mail/calendar-updates and **never** on security alerts, invites, or guard escalations before anything is flipped.
- **Phase 1 (`AWARENESS_DISPOSITION_GATE=on`, default off):** suppresses the approval row.
- Phase 2 (default-on) is a later flip after Phase 0 verification.

Promotion is inference-confidence-based (`twin-service.ts:619`), not raw approval-count, so suppressing awareness FYIs doesn't starve observer→suggest promotion.

## Tests
17 unit tests on `isAwarenessOnly` (every awareness tier, the human-mail exclusion, the injection-guard carve-out, each passive-shape guard, the flag default) + full `apps/api` suite green (913 passed).


jayzalowitz:
Addressed the one substantive review finding (the rest were verified-safe carve-outs): suppressing only the approval row left `decision_outcomes.requires_approval=true`, so the digest's `needsYou()` (`live-digest.ts:351`) still counted the item as a To-do — the two surfaces disagreed and the "needs you" count didn't drop. Fixed in `9a706f6`: when the gate fires we now flip the persisted outcome to `requires_approval=false` and re-save (`saveOutcome` upserts), so it's suppressed from the approval queue **and** bucketed as FYI in the digest. Also added the defensive `costZeroIntent==='unknown'` carve-out. Full apps/api suite green (914).

jayzalowitz:
All four review comments addressed:
1. **`costZeroIntent` guard (awareness-disposition.ts):** added in `9a706f6` — `isAwarenessOnly` now returns false when `costZeroIntent === 'unknown'` (the cost gate escalates those without a `confirmationLevel`).
2. **`requires_approval` left true → digest still counts it (events.ts):** this was the central bug — fixed in `9a706f6`. When the gate fires we flip the persisted outcome to `requires_approval=false` and re-save (`saveOutcome` upserts), so `needsYou()` buckets it as FYI **and** the approval row/SSE are skipped. The two surfaces now agree.
3. **No route-level test (events.ts):** added in `d765e17` — two `events-routes.test.ts` cases: flag-on ⇒ no approval row, no `approval:new`, outcome re-saved with `requires_approval=false`, `approval:null`; flag-off ⇒ approval still created (Phase 0 no-op).
4. **`costZeroIntent: 'unknown'` test (awareness-disposition.test.ts):** added in `9a706f6`.

Verified: 18 unit + 2 route tests; full apps/api suite green across 3 runs (916 passed).

PR #569, opened 2026-06-29
Stop reply-to-newsletter/no-reply in the core decision path

jayzalowitz:
## What
The reply-to-noise bug class — still open in the **primary** events→decision path. PRs #564/#568 only fixed the memory-suggestion path; live audit confirmed the core path was still vulnerable (latent on the account, but the code path is real).

## Root cause (verified)
- `inferEmailType` (gmail + outlook connectors) returns `meeting_invite` for any "meeting/invite/calendar" subject **before** checking the sender (`gmail-connector.ts:489` vs the no-reply check at `:498`). A webinar blast from `events-noreply@brand.com` → `requiresResponse: true`.
- `generateEmailTriageCandidates` (`decision-maker.ts:612`) generates `send_reply` purely on `requiresResponse` — it never reads the #251 authoring tier.

## Fix (two layers)
1. **Connectors**: an automated/no-reply sender can no longer become a reply-implying type (`meeting_invite`/`work_email`). Reuses `isAutomatedSender` from #568 (catches compound aliases like `google-noreply@`, `mailer-daemon@`).
2. **DecisionMaker**: suppress the `send_reply` candidate when the authoring tier ∈ awareness tiers (`inbox_newsletter`/`inbox_automated`/`user_sent_*`), even if `requiresResponse` slipped through — a robust gate that consults the specific signal, not the upstream heuristic.

Human inbound mail (`inbox_personal`/`inbox_broadcast`) is unaffected and still gets a reply offer.

## Tests
- connector: automated sender + meeting subject → `notification`; a real person + same subject → `meeting_invite`.
- decision-maker: reply offered for `inbox_personal`; never for `inbox_newsletter`/`inbox_automated`/`user_sent_originated`; tier honored top-level and nested under `data`.
- Full suites green: connectors 196, decision-engine 333.


jayzalowitz:
Addressed both review comments:
1. **isAutomatedSender called twice** — both connectors now compute `isNoReplySender` once and reuse it for the meeting gate + notification fallback (commit 047fade).
2. **No Outlook test** — added `OutlookMailConnector.inferEmailType` coverage (automated sender + meeting subject → notification; real person → meeting_invite), mirroring Gmail (commit ece0051).

PR #573, opened 2026-06-29
Honor chat calendar intent + harden escalate_to_user into a non-executing terminal

jayzalowitz:
## What

The intent classifier already tags chat-originated calendar decisions with `rawData.intent` — `decline_event` for "decline that meeting", `create_event` for "schedule a meeting with Sam". But `DecisionMaker.dispatchCandidates` keyed only off `situationType`, so the intent was **dead data**:

- "decline that meeting" routes to `calendar_update` → the generic menu is *acknowledge / dismiss* only. The decline the user explicitly asked for was never offered.
- "schedule a meeting" routes to `calendar_invite` → the menu is accept / tentative / decline of an inbound invite. But there is no invite — it carries an undefined `eventId` — and there's no *create* candidate at all.

This is the same root cause as PR1 (#569): dispatch keys off a coarse signal and discards the specific one.

## Change

- `generateCalendarUpdateCandidates`: on a chat `decline_event` intent, **lead with the `decline_invite`** the user asked for (HIGH confidence). acknowledge/dismiss stay as lower-confidence alternatives.
- `generateCalendarInviteCandidates`: on a chat `create_event` intent, return a single **`create_calendar_event`** candidate instead of the RSVP menu.
- Both branches gate on `rawData.source === 'chat'`, so a **real inbound** `calendar_update` / `calendar_invite` signal keeps its original menu unchanged.
- Both new candidates are genuine calendar writes (`calendar.events`), so the existing scope gate still downgrades them to a grant-access escalation when the scope isn't connected — verified by a test.

## Tests (decision-engine: 28 files, 338 pass)

- chat `decline_event` → leads with `decline_invite` (HIGH), acknowledge/dismiss still present
- the chat decline is scope-gated (downgraded to `escalate_to_user` without `calendar.events`)
- a **non-chat** `calendar_update` carrying an intent-like field keeps acknowledge/dismiss only (guards the `source==='chat'` gate)
- chat `create_event` → exactly one `create_calendar_event`, no accept/decline

## Notes / follow-ups

- Resolving "that meeting" → a concrete `eventId`, and parsing the event title/time out of "schedule a meeting with Sam at 3pm", are intentionally out of scope. This PR fixes the discarded-intent bug (surfacing the right action); the decline carries whatever `eventId` the upstream classifier resolved (today: possibly none → the action requires approval anyway, being irreversible).
- Branches off main at 0.6.88.0; VERSION → 0.6.89.0.


jayzalowitz:
## Cross-model review resolution (Claude adversarial + Codex)

Both adversarial passes converged on the same root issue, which reshaped this PR from a decision-maker tweak into a safety fix. Documenting how each finding was addressed:

**Claude subagent — both new candidate types dead-end at execution.** Confirmed: `create_calendar_event` has no handler in `CalendarActionHandler` (routes to `GenericActionHandler` → `success:false`), and a chat `decline_invite` carries no `eventId` (the classifier never sets one), which `CalendarActionHandler.execute` hard-fails with "Missing eventId". → **Fixed:** reframed both branches to a HIGH-confidence `escalate_to_user` (human-completion) instead of an auto-action. The intent is still recognized (the original bug — acknowledge/dismiss for a decline ask — is gone), but the user is asked to confirm rather than an action silently failing. Entity extraction + a create handler are the noted follow-up for full autonomy.

**Codex — `escalate_to_user` is not actually a guaranteed non-executing terminal.** This was the sharper, more important finding. `PolicyEvaluator` only force-approved `escalate_to_user` when `reason==='missing_write_scope'`; for every other escalation the "never auto-executes" guarantee was *incidental* to inbound signals carrying `untrusted_external` provenance. On a **trusted** path (a user's own chat intent, `user_originated`), a HIGH-confidence reversible zero-cost `escalate_to_user` clears `shouldAutoExecute` at `LOW_AUTONOMY`+, and the `events.ts` auto-execute branch routes it to the execution router (it's a registered OpenClaw action) → dead-ends. Safety Invariant #8 (security alerts never auto-execute) depended on this implicitly. → **Fixed:** `PolicyEvaluator.evaluate` now forces `requiresApproval` for **all** `escalate_to_user` regardless of tier/risk/provenance, so `autoExecute` is always false and it lands in the approval queue (already rendered as a "tell me what to do" card via `approvals.ts:225`). New policy tests prove it's forced to approval at `HIGH_AUTONOMY` where the same-shaped `archive_email` auto-executes.

**Forged-`source` note (Codex):** `/api/events/ingest` is behind `sessionAuth + requireOwnership`, and the worker forwarder overwrites `source` with the connector's real value, so inbound email content can't reach `rawData.source`. With the policy fix above, even a same-user forged `source:'chat'` escalation can only ever reach the approval queue, never auto-execute.

Tests: decision-engine 28 files / 338; policy-engine 14 / 193; evals 14 / 90 (incl. `safety-invariants-e2e`, `calendar-conflict-e2e`) — all green.

PR #577, opened 2026-06-29
idle-miner: ship SnapshotFileStore (device-local repos) + correct desktop-integration spec

jayzalowitz:
## What

Continuing the idle-miner build. Implementing it surfaced that the spec (#576) had the **central placement wrong**, and that the package was missing the one piece that made wiring hard everywhere. This PR fixes both: ships the load-bearing reusable building block, and corrects the spec with the architecture finding — without building into the wrong place.

## The architecture finding (building revealed it)

The spec assumed idle-miner runs in the Electron main process. Two facts say otherwise:

1. **`apps/desktop` is a deliberate thin shell — zero `@skytwin/*` deps, CommonJS** (`package.json` has no workspace deps; `tsconfig` is `module: commonjs`). It manages api/worker/cockroach as **child processes** over HTTP/IPC. Importing an ESM `@skytwin` package into Electron-main breaks that boundary.
2. **The worker is PAUSED exactly when idle-miner would run.** `IdlePauseController` (#382) calls `ServiceManager.pause()` (stops the worker child process) on idle. So the worker can't host the miner during idle either.

→ **Corrected host: a separately-managed idle-miner child process**, spawned by `ServiceManager` like api/worker/cockroach, started on the desktop's idle signal, exempt from the worker's idle-pause. Spec doc updated with this (`docs/idle-miner-desktop-integration.md`) — the original section is kept and marked superseded so the reasoning is visible.

## The reusable piece: `SnapshotFileStore`

`MinerOptions` requires a `fileIndexRepo` and a `cursorRepo`, but the package shipped **no implementation** — every host had to write one before it could run the miner, a big reason nothing ever wired it. This adds the sensible default:

- Implements both `FileIndexRepo` + `CursorRepo` over a single JSON snapshot under an injected dir.
- **Device-local, not CockroachDB** — a file index is per-machine state; a shared index would let one paired device suppress scans on another.
- **Persistence is load-bearing, not an optimization** — the signal pipeline has no content dedup, so an in-memory-only repo re-emits every scanned file on each restart.
- Atomic flush (write-temp + rename) on debounce + `close()` → a crash can never corrupt the index (worst case: re-scan the last unflushed window).
- Host-agnostic: correct regardless of where the miner is ultimately hosted.

## Tests (idle-miner: 6 files, 57 — incl. 10 new)

`snapshot-store.test.ts`: round-trip (index + cursor); **cross-instance persistence** (the no-re-emit-on-restart property) and its converse (unflushed mutations don't leak); key-collision safety for distinct `(rootId, relativePath)` pairs; atomic write (no lingering `.tmp`, valid JSON); corrupt-snapshot and version-mismatch → start fresh without throwing; `close()` flushes; no-op flush when unchanged.

## Scope

This retires the load-bearing risk (persistence correctness) and fixes the design, leaving the managed-process host + emitter-transform + userId + flag as the remaining wiring (now correctly targeted). Honest, tested, completable increment rather than a wrong-architecture build.


jayzalowitz:

Codex flagged five issues on `SnapshotFileStore`; three fixed, two documented-as-intended:

**Fixed (commit pushed):**
1. **Multi-writer `.tmp` race** — two stores sharing a dir could clobber each other's temp file and rename a half-written snapshot into place. → instance-unique temp name (`pid` + counter).
2. **Version-mismatch silently overwrote a newer snapshot** — on downgrade, a future-version snapshot was treated as empty, then the next flush wrote v1 over it, losing the data. → the incompatible snapshot is now **quarantined** to `<snapshot>.vN.bak` before starting fresh. Test added.
3. **Returned/stored-by-reference aliasing** — a caller mutating an object it passed to `upsert`/`save` or got back from `lookup`/`load` could diverge in-memory state from the recorded calls. → defensive shallow-copy on all four boundary methods (entries are flat). Test added.

**Documented as deliberate (not fixed):**
4. **No fsync** — `rename()` gives atomic visibility, not power-loss durability. Intentional: the file index is fully reconstructible, so a lost flush costs a re-scan of that window — cheaper than fsync latency on every flush. Now stated in an inline comment.
5. **Full-snapshot serialization / unbounded growth** — bounded by scanned-file count; the debounce coalesces a burst into one write per window, so churn doesn't flush per-file. For very large trees a SQLite-backed repo is the documented alternative (`docs/idle-miner-desktop-integration.md`). Now noted inline.

Tests: idle-miner 6 files / 58 (11 on the store).

PR #578, opened 2026-06-29
idle-miner: ship the signal emitter (RawSignal→ingest transform + HTTP delivery)

jayzalowitz:
## What

Continuing the idle-miner reusable core. After the repos (`SnapshotFileStore`, #577), the other `MinerOptions` dependency the package never shipped is the `signalEmitter` — so every host had to hand-write the filesystem-`RawSignal` → ingest mapping. This ships it.

## Change

`packages/idle-miner/src/ingest-adapter.ts`:

- **`toIngestEvent(signal, userId)`** — pure transform. idle-miner's filesystem `RawSignal` (`rootId/absPath/relPath/structuredFields`) is a *different shape* than the connector `RawSignal` the worker forwards, and `/api/events/ingest` dispatches on `source`/`type`, so this wraps it in that envelope with `source: 'fs'` (→ `kind: 'fs'` in capability inference).
  - **Security:** the file-derived `structuredFields` (a package.json `name`, a git remote, …) are only semi-trusted, so they're spread **first** — the trusted envelope fields (`source` / `type` / `signalId` / `userId`) always win. A file can't set `source: 'gmail'` or re-attribute the signal to another `userId` by naming an extracted field after an envelope key. Covered by a test.
- **`createHttpSignalEmitter({ ingestUrl, userId })`** — returns the `MinerOptions.signalEmitter` shape. POSTs the event with bounded exponential-backoff retry on transient (429 / 5xx / network) failures via `@skytwin/core`'s `withRetry`, and **never throws into the miner loop**: a final failure is logged and swallowed, because the durable file index re-attempts the signal on a later scan (crashing the whole scan over one failed POST would be strictly worse).

## Tests (idle-miner: 7 files, 66 — incl. 8 new)

`ingest-adapter.test.ts`: envelope mapping; `file_skipped` vs `file_indexed`; the **structuredFields-can't-override-envelope** security property; optional-field omission; emitter POSTs the right body/URL; swallows a permanent 4xx (not retried); swallows a network error; retries a transient 503 then succeeds.

## Where this leaves the miner

The reusable core is now **complete and package-provided**: `SnapshotFileStore` (repos) + `createHttpSignalEmitter` (delivery) + `DEFAULT_EXTRACTORS` + `expandAllowlist` (roots) + `ElectronIdleDetector`. The only work left to actually run the miner is host-specific: the separately-managed child process (per the #577 architecture correction), the resolved `userId`, and the `SKYTWIN_IDLE_MINER_ENABLED` flag. Spec (`docs/idle-miner-desktop-integration.md`) updated.


jayzalowitz:

Codex flagged four issues on the ingest adapter; all addressed:

1. **`structuredFields` could inject arbitrary top-level keys** (envelope keys were protected from shadowing, but other file-derived keys — and a `__proto__` in the serialized JSON — still reached the top level). → File-derived metadata is now **nested under a single `extracted` key**; the top level is fully controlled by `toIngestEvent`. Test asserts no shadow + no top-level injection.
2. **`408 Request Timeout` was treated as permanent** (not retried). → Added to the retryable set alongside 429/5xx. Test added.
3. **No bound on extracted-metadata size** → oversized `extracted` is now **dropped** (the miner emits bounded metadata by design — a name, a git remote, never file content — so this only fires on pathological input). Test added.
4. **Permanent config failures swallowed at `warn` forever** (wrong URL / auth / 4xx indistinguishable from transient). → permanent failures are thrown as a `PermanentIngestError` and logged at **error** severity ("check ingestUrl / auth / payload schema"); transient drops stay at `warn`. Test asserts a permanent 4xx is not retried.

Tests: idle-miner 7 files / 70 (12 on the adapter).

PR #590, opened 2026-07-05
Build apps/idle-miner-runner — the managed idle-miner process (+ EventDrivenIdleDetector)

jayzalowitz:
## What

The idle-miner reusable core is package-complete (repos #577, emitter #578). This ships the **managed process** that assembles and runs it — the thing the desktop `ServiceManager` will spawn.

Per the #577 architecture correction, idle-miner can't run in the Electron main process (a CJS shell with zero `@skytwin` deps) or in the worker (paused exactly when idle-miner would run). So it's a separately-managed ESM child process, spawned like api/worker/cockroach.

## Change

**`apps/idle-miner-runner`** (new app):
- **`config.ts`** — parses the env the `ServiceManager` passes, **fail-closed** (it scans the user's real filesystem, so no ambiguous defaults): `SKYTWIN_IDLE_MINER_ENABLED` must be exactly `true` (**default off**, defense-in-depth with the desktop's spawn gate); `SKYTWIN_IDLE_MINER_USER_ID` / `_INGEST_URL` (URL-validated) / `_DATA_DIR` required; home from `SKYTWIN_IDLE_MINER_HOME` / `HOME` / `USERPROFILE`.
- **`runner.ts`** — assembles the miner entirely from the package core (`SnapshotFileStore` at the data dir, `createHttpSignalEmitter` at the ingest URL, `expandAllowlist`→`FsScanRoot[]`, `DEFAULT_EXTRACTORS`), driven by the new `EventDrivenIdleDetector`. Returns a small control surface (`handleControlLine`, `shutdown`).
- **`index.ts`** — thin glue: parse → fail-closed exit / assemble → read `idle`/`active`/`stop` control words the parent writes to **stdin** → clean shutdown on SIGTERM/SIGINT/stdin-close (flushes the device-local index).

**`EventDrivenIdleDetector`** (new, in `@skytwin/idle-miner`): a plain Node child has no Electron `powerMonitor`, so the process can't self-detect idle. The parent (which owns OS idle detection) drives transitions via `setIdle()`/`setActive()`; handlers fire only while started and same-state signals are de-duped to clean edges.

## Tests (26 new)

- runner app (21): fail-closed config across the flag (default off, only `"true"` enables) + every required field + malformed URL + home fallbacks incl. `USERPROFILE`; assembly wires the allowlisted roots / injected detector / store-as-both-repos; the control protocol (case/whitespace-insensitive, ignores unknown lines); idempotent shutdown that stops the miner, flushes the store, and halts the detector.
- `EventDrivenIdleDetector` (5): no fire before `start()`; clean idle↔active edges; de-dup of repeated signals; `stop()` halts relays.

## Remaining (desktop-side, the last step)

`ServiceManager` spawning this process + piping `IdleBridge` transitions to its stdin + embedded-build packaging + resolving the paired `userId` (persist to a desktop pref on pair). Spec (`docs/idle-miner-desktop-integration.md`) updated.


jayzalowitz:

Three findings; two fixed, one dismissed as consistent-with-worker:

1. **HIGH — shutdown not drained.** `handle.stop()` only flipped stop flags; an in-flight scan batch could be mid-file/HTTP/write while `store.close()` flushed and the process exited, losing work. → Added **`IdleMinerHandle.drain()`** (awaits the in-flight `activeBatch` after `stop()`); `runner.shutdown()` is now async and drains **before** flushing the index; the entrypoint awaits it before `process.exit`. Tests assert drain-before-flush ordering + that drain awaits an in-flight batch.
2. **MEDIUM — `stop` control word didn't exit the process.** It stopped the miner but left a live child (stdin open, detector stopped). → The entrypoint (`index.ts`) now owns process exit: `stop` / SIGTERM / SIGINT / stdin-close → async app-shutdown → `exit(0)`. The runner no longer handles `stop` (test updated to assert it's inert at the runner level).
3. **HIGH — emitter posts unauthenticated.** Verified this is **consistent with the worker**, not a new gap: `/api/events` is behind `sessionAuth` + `requireOwnership`, but the worker's `forwardSignalToApi` — the *primary* signal path — also posts unauthenticated and works because the desktop is a single-user local install with the API's dev/localhost auth bypass (`SKYTWIN_DEV_AUTH_BYPASS` + localhost). This process runs on the same host and posts from localhost identically. A hardened multi-tenant deploy would give **both** the worker and this process a service credential — a shared concern for the local-forwarding architecture, not specific to idle mining. Documented inline in `runner.ts`.

Tests: idle-miner 9 files / 77 (incl. 2 new drain), runner 2 / 21.

PR #594, opened 2026-07-05
feat(routines): worker scheduler — watches now fire (#519 pt.3b)

jayzalowitz:
## What

Part 3b of no-code routines (#519): the **gated worker scheduler** that makes Watches actually fire. It consumes the part-3a engine (`computeNextRun` + `matchesFilter`) and the part-2 storage.

Each minute (opt-in, `SKYTWIN_WATCHES_ENABLED`, default off): claim every **due** watch (`listDue`), match the user's recent signals against its filter, record the result, and schedule the next firing in the user's timezone.

## This PR

- **db:** migration `070-watch-runs` + `watchRunRepository`. A `watch_run` is the **canonical, explanation-carrying record** of a firing: `matched_refs` = the evidence (which signals matched), `summary` = the digest/notify output, and the watch's own `filter` (joinable via `watch_id`) = the "why" — the read-only equivalent of an `ExplanationRecord`. The ambient surfaces (briefing projection, notifications — part 4) will READ from this; single source of truth, no separate copy.
- **worker:** `watch-scheduler` job — a pure gate (`shouldRunWatchScheduler`), pure matcher/summary (`evaluateWatch`/`toMatchable`), and the runner: per due watch it loads the tz, windows signals by last run (cadence-sized first-run lookback so a digest never re-summarizes the same signals), `matchesFilter`, writes a `watch_run` **only on ≥1 match**, and **always** `markRan(computeNextRun)` to advance the schedule. Wired into the poll loop like the other opt-in jobs.

## Safety

**Read-only, no policy gate** — a firing produces a digest/notification and executes nothing; the `watch_runs.action` CHECK constraint enforces `digest`/`notify`. Each watch is isolated in its own try/catch so one failure (bad tz, signal-query error) can't stall the rest.

## Verification

Migration DDL **verified against real CockroachDB** (table, indexes, action CHECK rejecting a non-read-only action). 11 scheduler unit tests (gate; matcher/window; digest vs notify summary; runner writes-a-run, advances-on-no-match, failure-isolation). db 453 / worker 172 green; full build green.

Next: **part 4** surfaces the runs — the briefing projection (ambient), the Watches management page (run history), and chat authoring.


jayzalowitz:
## Review resolution (adversarial /review + Copilot)

Pushed two fix commits (`ceb7884`, `b115e33`).

### Adversarial /review (cross-model) — all addressed
- **Critical: claim-before-process could drop a window on crash.** Reordered to **evaluate-then-claim** — the scheduler fetches + evaluates a watch's signals *before* `claimDue` advances `next_run_at`. A crash/transient failure during the expensive fetch no longer permanently loses the firing; it retries next tick. The atomic claim still guarantees exactly-once across workers (the loser discards its computed run).
- **High: fresh watch re-summarized pre-creation signals.** `windowStart` is now floored at the watch's `created_at`.
- **High: unbounded per-tick work after long downtime.** `windowStart` capped at a 7-day ceiling (`MAX_WINDOW_MS`), bounding the `getRecent` lookback.
- **Medium: boundary signal double-count/miss.** The window is now the half-open interval `(windowStart, claimTime]` with an explicit upper bound; a signal at the exact boundary is counted once. `matched_refs` capped at 200 while `matched_count` keeps the true total.
- **Migration:** added `CHECK (matched_count >= 0)` — verified against real CockroachDB (SQLSTATE 23514 on a negative insert).

### Copilot — all 5 addressed
1–3. `watch-run-repository.ts` type nits: `matched_refs` typed non-null (matches `NOT NULL DEFAULT '[]'`), `action` typed `RoutineActionKind` (matches the CHECK), redundant `?? []` dropped, `listRecentForUser(since: Date)` instead of a string.
4. **`listDue` plain-read double-fires across workers** — resolved by the same-PR `claimDue` optimistic claim (atomic `next_run_at` transition gated on the seen value).
5. **Unbounded `getRecent` lookback** — resolved by the `MAX_WINDOW_MS` window ceiling (lookback ≤ ~169h).

16 unit cases green; db + worker build clean. Re-requesting Copilot for a fresh pass on the reordered runner.

